PR #1, opened 2020-08-04
Add rust functions for set_thread_name in the Tracy C API

zanshi:
Tracy now has support for setting the thread name in the C API, this PR adds support for this!

I think you might have to publish the new versions to crates.io for everything to work, I'm not really sure how that works though.

I also ran `cargo fmt`, hence the formatting changes.

### DIFF
--- a/tracy-client-sys/Cargo.toml
+++ b/tracy-client-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tracy-client-sys"
-version = "0.8.0"
+version = "0.8.1"
 authors = ["Simonas Kazlauskas <tracy-client-sys@kazlauskas.me>"]
 build = "build.rs"
 license = "MIT/Apache-2.0"
--- a/tracy-client-sys/src/lib.rs
+++ b/tracy-client-sys/src/lib.rs
@@ -165,6 +165,10 @@ enabled_fn! { pub fn ___tracy_emit_message_appinfo(
     size: usize
 ) }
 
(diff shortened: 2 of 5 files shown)


nagisa:
Very cool. I’ll see if I can get the CI to pass without publishing things to crates.io first.

nagisa:
Thanks!

lqd:
Is tracy's `set_thread_name` API actually released ? Otherwise, this client would now stop being compatible with tracy 0.7 for example.

nagisa:
The C++ API that the function ultimately calls [did exist in v0.7.0](https://github.com/wolfpld/tracy/blob/v0.7/common/TracySystem.cpp#L101), and is briefly documented in the section 2.4 of the manual. This API was not bound for C API clients. That happened [more recently](https://github.com/wolfpld/tracy/commit/826affb0b557474eba3335dc2a21668ea3c0f136). Since the new C API is just binding to the pre-existing C++ API, there's no reason why it wouldn't work.

lqd:
I tried the client to make sure and when connecting the GUI on version 0.7, it errors saying the protocol is incompatible. It might not be the function per se causing this, rather another commit being pulled in with the submodule update. Or maybe the error is just in my setup if it's working fine for you.

nagisa:
I see. That's a shame, I'll yank the release then.

lqd:
maybe test it first on your side in case it really is just an issue on my machine ?

nagisa:
Could repro, and yanked. Will have to wait for the next tracy release before this functionality is made available.

PR #5, opened 2020-09-14
Add support for optionally using externally built Tracy library

zanshi:
This is useful in scenarios with multi-dll/shared objects, e.g. when you have a C++ project with a Rust component and you want to have the Tracy profiling in a single view. 

I ran `cargo fmt` on all the projects as well.

I'm not sure about the name of the flag really, do you have any suggestions? I mean, all it does is disable compilation of Tracy, it doesn't actually try to link to an externally built library.

Sorry for the previous PR with the backwards comp breakage and so on... This one just adds an optional feature flag :)


nagisa:
I’m somewhat uncomfortable with this, mostly because `tracy` is not very good at maintaining a stable ABI and versioning the project accordingly.

zanshi:
> I’m somewhat uncomfortable with this, mostly because `tracy` is not very good at maintaining a stable ABI and versioning the project accordingly.

Even if the default behaviour of the crate remains the same?

I agree that the ABI changes constantly, even between releases, but I think this is a useful feature for people who just want the function definitions that `tracy-client-sys` provides, without building the library.

nagisa:
> Even if the default behaviour of the crate remains the same?

Well, it still exposes a way for people to shoot themselves in a foot, by means that are not immediately apparent as to their severity.

For `tracy-client-sys` to support the use-case of using arbitrary version of tracy it would need to be rewritten to generate the bindings at build time from the provided headers (via bindgen), which by itself is a not a terrible proposition. At the very least it would make any mismatches explode loudly at compile time. With the way `tracy-client-sys` bindings are hand-written currently, were we to allow users to provide an arbitrary version of `tracy` binary code, the signature mismatches would compile correctly, but then clobber registers, memory etc at runtime, leading to difficult to debug crashes etc.

zanshi:
> > Even if the default behaviour of the crate remains the same?
> 
> Well, it still exposes a way for people to shoot themselves in a foot, by means that are not immediately apparent as to their severity.
> 
> For `tracy-client-sys` to support the use-case of using arbitrary version of tracy it would need to be rewritten to generate the bindings at build time from the provided headers (via bindgen), which by itself is a not a terrible proposition. At the very least it would make any mismatches explode loudly at compile time. With the way `tracy-client-sys` bindings are hand-written currently, were we to allow users to provide an arbitrary version of `tracy` binary code, the signature mismatches would compile correctly, but then clobber registers, memory etc at runtime, leading to difficult to debug crashes etc.

Hmm, I see your point. I think I'll keep my own branch with the changes for now. I like the `bindgen` idea! I might look into that in the future :)

PR #7, opened 2020-11-04
Add all named fields to the span name

superdump:
This makes spans much more useful for me so I can see all the additional metadata in Tracy. Particularly having a span in bevy with a name of `"system"` and a field like `name = system.name()`, with this PR I can tell what systems are what in Tracy. I'm open to input for how to improve it.

### DIFF
--- a/tracing-tracy/src/lib.rs
+++ b/tracing-tracy/src/lib.rs
@@ -212,6 +212,14 @@ mod tests {
     }
 
     #[test]
+    fn with_fields() {
+        setup_subscriber();
+        let span = span!(Level::TRACE, "a sec", name = "test");
+        let _enter = span.enter();
+        event!(Level::INFO, "EXPLOSION IN A SPAN WITH A NAME!");
+    }
+
+    #[test]
     fn multiple_entries() {
         setup_subscriber();
         let span = span!(Level::INFO, "multiple_entries");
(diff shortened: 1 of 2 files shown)


nagisa:
Can you please add a test that makes sure both branches are exercised?

superdump:
Done. I ran `cargo fmt` as well.

nagisa:
I won't be able to look at/verify/check this for a couple more days, but I'm keeping the tab open as a reminder. If I don't get back to this please ping me.

tracing-tracy/src/lib.rs
```diff
@@ -80,19 +86,25 @@ impl Default for TracyLayer {
...
 {
     fn on_enter(&self, id: &Id, ctx: Context<S>) {
         if let Some(span_data) = ctx.span(id) {
             let metadata = span_data.metadata();
             let file = metadata.file().unwrap_or("<error: not available>");
             let line = metadata.line().unwrap_or(0);
+            let name = if let Some(fields) = span_data.extensions().get::<FormattedFields<F>>() {
+                format!("{}: {}", metadata.name(), fields.fields.as_str())
+            } else {
+                metadata.name().to_string()
```
nagisa, line 102: Consider using a `Cow<'_, str>` here, perhaps? It would also make sense to me to lift and duplicate the `Span::new` call into the conditional here.

nagisa (commented):
I tried running the newly added tests locally and checked them in Tracy, but I'm failing to see the change in behaviour I would intuitively expect from this PR. Am I missing something?

nagisa:
This has been adapted and merged into main branch. Will be released shortly.

PR #16, opened 2021-05-13
Expose ondemand feature

ibab:
This exposes the TRACY_ON_DEMAND feature (don't record traces until there is a connection, and allow re-connecting to a client).
I had to make a small change to `make_sys.sh` to make it work on MacOS.


tracy-client/Cargo.toml
```diff
@@ -21,3 +21,4 @@ enable = [ "tracy-client-sys/enable" ]
 delayed-init = [ "tracy-client-sys/delayed-init" ]
 lowres-timer = [ "tracy-client-sys/lowres-timer" ]
 noexit = [ "tracy-client-sys/noexit" ]
+ondemand = [ "tracy-client-sys/ondemand" ]
```
nagisa, line 24: Will want to add a similar one to `tracing-tracy`.

nagisa:
Note that the `make_sys` is still being developed (while I have time) so it is likely that it won't work on e.g. macOS. It is intended to be only run by the GHA, so the changes there aren't necessary. Would be great if this PR only restricted itself to just the `on_demand` feature.

nagisa:
I went ahead and merged the two commits adding the feature manually. Thanks!

ibab:
Thank you!

PR #21, opened 2021-09-06
Add span! macro

benfrankel:
The macro uses `file!()` and `line!()` to determine the file name and line number to pass to `Span::new(...)`. It also uses a hack to determine the function name, which is the same hack used in [`stdext::function_name!`](https://docs.rs/stdext/0.3.1/src/stdext/macros.rs.html#63-74), since the [RFC](https://github.com/rust-lang/rfcs/issues/1743) to add a macro for this to the Rust stdlib is unresolved.

Alternatively, `stdext` could be added as a dependency, but that's probably overkill.

Also, it might be convenient to support omitting the `callstack_depth` with a default value of 62, which is the max according to Tracy documentation ("The maximum call stack depth that can be retrieved is 62 frames. This is a restriction at the level of operating system.").

### DIFF
--- a/tracy-client/src/lib.rs
+++ b/tracy-client/src/lib.rs
@@ -19,6 +19,27 @@ use std::alloc;
 use std::ffi::CString;
 use tracy_client_sys as sys;
 
+/// Start a new Tracy span with function, file, and line determined automatically.
+#[macro_export]
+macro_rules! span {
+    ($name:expr, $callstack_depth:expr) => {{
+        fn f() {}
+        fn type_name_of<T>(_: T) -> &'static str {
+            std::any::type_name::<T>()
+        }
+        let func_name = type_name_of(f);
+        let function = &func_name[..func_name.len() - 3];
+
+        tracy_client::Span::new(
+            $name,
+            function,
+            file!(),
+            line!(),
+            $callstack_depth,
+        )
+    }};
+}
+
 /// A handle representing a span of execution.
 #[cfg(feature="enable")]
 pub struct Span(


Ralith:
This would be awesome to have!

> Also, it might be convenient to support omitting the callstack_depth with a default value of 62

Sounds like a good plan to me.

tracy-client/src/lib.rs
```diff
@@ -19,6 +19,27 @@ use std::alloc;
...
 use tracy_client_sys as sys;
 
+/// Start a new Tracy span with function, file, and line determined automatically.
+#[macro_export]
+macro_rules! span {
+    ($name:expr, $callstack_depth:expr) => {{
+        fn f() {}
+        fn type_name_of<T>(_: T) -> &'static str {
+            std::any::type_name::<T>()
+        }
```
nagisa, line 29: From what I can tell this can be simplified to a

```rust
struct S; 
let func_name = std::any::type_name::<S>();
let function = &func_name[..func_name.len() - 3];
```

nagisa (approved):
This seems reasonable to me!

Sorry for taking a while to get to this.

nagisa:
Would be great if you could add a test that just calls the macro as well as bump the version of rustc in CI files to something like 1.52 or 1.53.

nagisa:
Merged in https://github.com/nagisa/rust_tracy_client/commit/f06d0fdd246fb0c1a8d8a25e246f5080164cf16d

Thanks!

Ralith:
Awesome! Any chance of a patch release with this in it soon?

nagisa:
Yeah I have been meaning to make some other changes too but I guess i can
just split them up.

nagisa:
Looks like we're seeing some SIGSEGVs on Windows in a doctest I added for the macro, which I'll need to investigate before I cut a release.

nagisa:
Could not reproduce the issue locally and I bet its not a new failure mode anyway, so went ahead and published 0.12.5.

PR #23, opened 2021-10-19
Relax literal the requirement of the create_plot macro so that it can be used with stringify

Imberflur:
This matches what `concat!` uses https://doc.rust-lang.org/stable/core/macro.concat.html

### DIFF
--- a/tracy-client/src/lib.rs
+++ b/tracy-client/src/lib.rs
@@ -341,7 +341,7 @@ pub fn set_thread_name(name: &str) {
 /// ```
 #[macro_export]
 macro_rules! create_plot {
-    ($name: literal) => {
+    ($name: expr) => {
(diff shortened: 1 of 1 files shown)


Imberflur:
Also, `concat!` will produce a friendly customized error if a literal is not used:
image
Compared too:
image
or with multiple macro layers:
image

nagisa:
I'm sorry, but I won't be able to take a careful look into this PR for a couple more weeks (currently undergoing a job change)

nagisa:
Actually, never mind. I'll merge this now and release a version with this in place and we can release a breaking change later down the line if this turns out to have been a mistake.

Imberflur:
@nagisa I'm happy to wait

nagisa:
tracy-client v0.12.6 is up.

PR #28, opened 2021-11-29
Add `static_span!` macro that logs without allocating

### LINKED ISSUE #26: Logging a span without allocating
Hello! I was profiling a very performance sensitive app, and I wanted to get tracing overhead as low as possible. The implementation of `Span` calls `___tracy_alloc_strloc_name` and `___tracy_emit_zone_begin_alloc_callstack`, which were too expensive for my usecase.

In my fork (https://github.com/sujayakar/rust_tracy_client), I've added a `static_span!` macro that avoids dynamic allocation by using `&'static str`s for the function name and file name. The implementation (https://github.com/sujayakar/rust_tracy_client/blob/main/tracy-client/src/lib.rs#L67) uses some pretty gnarly nightly features. Would you be interested in my upstreaming this?

sujayakar:
Sorry for the delay submitting this! This PR adds `static_span!`, gated behind a `static_span` feature gate since it depends on nightly features.

I verified that the crate works without nightly without the feature on
```
sujayakar@Sujays-MBP tracy-client % rustup override set stable
sujayakar@Sujays-MBP tracy-client % cargo test                       
```
and that the `const_format` dependency only shows up when the feature is on.
```
sujayakar@Sujays-MBP tracy-client % cargo tree                   
tracy-client v0.12.6 (/Users/sujayakar/src/rust_tracy_client/tracy-client)
└── tracy-client-sys v0.16.1 (/Users/sujayakar/src/rust_tracy_client/tracy-client-sys)
    [build-dependencies]
    └── cc v1.0.72
sujayakar@Sujays-MBP tracy-client % cargo tree --features static_span
tracy-client v0.12.6 (/Users/sujayakar/src/rust_tracy_client/tracy-client)
├── const_format v0.2.22
│   └── const_format_proc_macros v0.2.22 (proc-macro)
│       ├── proc-macro2 v1.0.32
│       │   └── unicode-xid v0.2.2
│       ├── quote v1.0.10
│       │   └── proc-macro2 v1.0.32 (*)
│       └── unicode-xid v0.2.2
└── tracy-client-sys v0.16.1 (/Users/sujayakar/src/rust_tracy_client/tracy-client-sys)
    [build-dependencies]
    └── cc v1.0.72
```

Closes #26.

### DIFF
--- a/tracy-client/src/lib.rs
+++ b/tracy-client/src/lib.rs
@@ -50,7 +50,14 @@ macro_rules! span {
     }};
 }
 
-#[cfg(feature="static_span")]
+#[cfg(all(feature="static_span", not(feature="enable")))]
+#[macro_export]
+macro_rules! static_span {
+    () => {{}};
+    ($name:expr) => {{}};
+}
+
+#[cfg(all(feature="static_span", feature="enable"))]
 #[macro_export]
 macro_rules! static_span {
     () => {{
@@ -169,16 +176,9 @@ impl Span {
     }
 
     #[doc(hidden)]
-    #[inline]
+    #[cfg(feature="enable")]
     pub fn private_new(ctx: sys::___tracy_c_zone_context) -> Self {
-        #[cfg(not(feature="enable"))]
-        {
-            return Self(());
-        }
-        #[cfg(feature="enable")]
-        {
-            Self(ctx, std::marker::PhantomData)
-        }
+        Self(ctx, std::marker::PhantomData)
     }
 
     /// Emit a numeric value associated with this span.
--- a/tracy-client/src/lib.rs
+++ b/tracy-client/src/lib.rs
@@ -14,10 +14,14 @@
 //!
 //! [Tracy profiler]: https://github.com/wolfpld/tracy
 #![cfg_attr(not(feature="enable"), allow(unused_imports, unused_variables))]
+#![cfg_attr(feature="static_span", feature(const_type_name, const_mut_refs))]
 
 use std::alloc;
 use std::ffi::CString;
 use tracy_client_sys as sys;
+
+#[cfg(feature="static_span")]
+pub use const_format as const_format;
 
 /// Start a new Tracy span with function, file, and line determined automatically.
 ///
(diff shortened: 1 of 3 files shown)


Ralith:
I would much rather this not be a separate macro. As discussed in #26, there are also approaches that would not require nightly.

sujayakar:
cool, so to make sure I understand the proposal, the idea is to have a `OnceCell<[u8; MAX_NAME_LEN]>` per callsite to `span!` and then lazily initialize it on the first log event. the initialization would mostly be just copying a `&'static str` over and adding the null terminator. 

for my microsecond scale application, I'd still like to avoid the overhead of synchronizing and checking the `OnceCell`. would you be open to having the nightly-only implementation I've submitted here under a "nightly" feature? that way users can opt into the faster implementation of `span!` that does all the work at compile time if they're okay with nightly.

then, with those changes, the only differences between `static_span` and `span` is the use of `tracy_emit_zone_begin` and `tracy_emit_zone_begin_callstack`. as with the current code, we can use `callstack_depth = 0` to know when to avoid collecting the callstack (which is also unacceptably slow for my application).

Ralith:
> `OnceCell<[u8; MAX_NAME_LEN]>`

I think we can just use `OnceCell<String>` (or `CString` if that's more convenient).

> would you be open to having the nightly-only implementation I've submitted here under a "nightly" feature?

I see no harm in swapping in a more efficient version behind the scenes when available, certainly!

tracy-client/src/lib.rs
```diff
@@ -46,6 +50,79 @@ macro_rules! span {
...
+            FILE,
+            "\0"
+        );
+
+        // Source: https://stackoverflow.com/questions/38088067/equivalent-of-func-or-function-in-rust
+        fn f() {}
+        const fn type_name_of<T>(_: &T) -> &'static str {
+            std::any::type_name::<T>()
+        }
+        const F_FN_NAME: &'static str = type_name_of(&f);
```
nagisa, line 79: The `span!` macro above has a terser version of this.

tracy-client/src/lib.rs
```diff
@@ -46,6 +50,79 @@ macro_rules! span {
...
+        const FN_NAME_C: &'static str = $crate::const_format::concatc!(
+            $crate::const_format::Sliced(F_FN_NAME, 0..F_FN_NAME.len() - 3),
+            "\0"
+        );
+        const SRC_LOC: $crate::sys::___tracy_source_location_data = $crate::sys::___tracy_source_location_data {
+            name: FN_NAME_C.as_ptr() as *const _,
```
nagisa, line 85: Not entirely sure a default such as this is the best. For instance, this would produce a pretty poor experience if a person used `static_span` twice in the same function.

nagisa:
Sorry for the late feedback. the `static_span` doing nothing when a feature is not enabled is pretty sketchy API design wise. To avoid that I think it would indeed make a bit more sense to make it a part of a regular span macro, potentially one which switches implementations based on availability. Ultimately we return a `Span` either way, don't we?

Ralith:
@sujayakar do you plan to come back to this?

sujayakar:
hey, sorry I've been busy with other stuff, and it's sitting in my backlog. we can close this out in the meantime if you'd prefer.

Ralith:
I believe this, and #26, can be closed due to #33.

sujayakar:
awesome! glad to see it got in with #33.

PR #33, opened 2022-04-03
Non-allocating `span!` macro

Ralith:
Alternative take on https://github.com/nagisa/rust_tracy_client/pull/28.

This introduces a new `once_cell` dependency; however, this is small, very widespread, and likely to move into std (https://github.com/rust-lang/rust/issues/74465).

Care was taken to ensure that the `Lazy` values are not dereferenced outside of the `#[cfg(feature = "enable")]` gate. Combined with inlining, this should allow invocations to be entirely erased when disabled, and very cheap when enabled.

Tests pass, but I haven't tried actually using this yet.

### DIFF
--- a/tracy-client/src/lib.rs
+++ b/tracy-client/src/lib.rs
@@ -31,19 +31,64 @@ use tracy_client_sys as sys;
 #[macro_export]
 macro_rules! span {
     ($name:expr) => {
-        $crate::span!($name, 62)
+        $crate::span_impl!($name, 62)
     };
+    ($name:expr, $callstack_depth:expr) => {
+        $crate::span_impl!($name, $callstack_depth)
+    };
+}
+
+#[macro_export]
+#[doc(hidden)]
+#[cfg(feature = "enable")]
+macro_rules! span_impl {
     ($name:expr, $callstack_depth:expr) => {{
+        use std::ffi::CString;
+        use $crate::internal::once_cell::sync::Lazy;
+
         struct S;
-        let func_name = std::any::type_name::<S>();
-        $crate::Span::new(
-            $name,
-            &func_name[..func_name.len() - 3],
-            file!(),
-            line!(),
-            $callstack_depth,
-        )
+        static SRC_LOC: Lazy<$crate::internal::SrcLoc> = Lazy::new(|| {
+            let function_name = std::any::type_name::<S>();
+            let function_name = CString::new(&function_name[..function_name.len() - 3]).unwrap();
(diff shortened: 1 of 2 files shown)


tracy-client/src/lib.rs
```diff
@@ -34,18 +37,33 @@ macro_rules! span {
...
-            $callstack_depth,
-        )
+        static FUNC_NAME: Lazy<CString> = Lazy::new(|| {
+            let func_name = std::any::type_name::<S>();
+            CString::new(&func_name[..func_name.len() - 3]).unwrap()
+        });
+        static SRC_LOC: Lazy<$crate::SrcLoc> = Lazy::new(|| {
+            $crate::SrcLoc($crate::sys::___tracy_source_location_data {
+                name: concat!($name, "\0").as_ptr().cast(),
+                function: FUNC_NAME.as_ptr(),
```
nagisa, line 51: We can probably avoid the two `Lazy`s per span site, right? You could probably just create the `CString` in this same `SRC_LOC` `Lazy`, grab the pointer and `std::mem::forget` the `CString` itself.

(And if the two `Lazy`s do in fact serve some purpose, a comment would be nice.)

tracy-client/src/lib.rs
```diff
@@ -34,18 +37,33 @@ macro_rules! span {
         $crate::span!($name, 62)
     };
     ($name:expr, $callstack_depth:expr) => {{
+        use std::ffi::CString;
```
nagisa, line 40: Should re-export `CString` from the `$crate` as well. I would prefer a `#[doc(hidden)] pub mod internal` with most of the the internal stuff hidden in there over hiding each separate item.

Ralith: Is renaming std a thing? `std::any::type_name` wasn't reexported previously.

nagisa: I kinda did that a couple of times for codebases that needed to make libstd support conditional. But I agree the `CString`/`type_name` reexport doesn't need to happen now or here.

tracy-client/src/lib.rs
Ralith: Refactored to use a single `Lazy`.

Ralith:
Tweaked things a bit to use a secondary macro rather than relying on optimizer magic. The macro's outside the `mod internal` because I don't think declarative macros can be exported from anywhere but the top level anyway.

nagisa:
Thanks!

PR #34, opened 2022-04-26
0.8.1 changes

### LINKED ISSUE #29: tracing_tracy segfaults when main thread exits with other threads still running
When the main thread exits, if there are other threads that are still running, closing a span in one of those threads appears to cause a segfault.  (The threads are stopped when the process exits, but it seems that they keep running while the main thread is exiting).

This example always causes a segfault on my machine (with tracy-client `0.12.6`)
```rust
fn main() {
    std::thread::spawn(move || {
        loop {
            let span = tracy_client::span!("Temp");
            drop(span);
        }
    });
    std::thread::sleep(std::time::Duration::from_millis(50));
}
```

Here's a backtrace from the core dump:
```
#0  memcpy (__len=1, __src=<synthetic pointer>, __dest=0x7f880042adc0) at /usr/include/x86_64-linux-gnu/bits/string_fortified.h:29
#1  tracy::MemWrite<tracy::QueueType> (val=tracy::QueueType::Callstack, ptr=0x7f880042adc0) at tracy/client/../common/TracyAlign.hpp:22
#2  tracy::Profiler::SendCallstack (this=<optimized out>, depth=depth@entry=60, skipBefore=skipBefore@entry=0x55c8703a0cf3 "__kernel_rt_sigreturn")
    at tracy/client/TracyProfiler.cpp:3058
#3  0x000055c870365499 in tracy::CrashHandler (signal=11, info=0x7f8802edb170) at tracy/client/TracyProfiler.cpp:900
#4  <signal handler called>
#5  0x000055c870366701 in memcpy (__len=<optimized out>, __src=<optimized out>, __dest=<optimized out>) at /usr/include/x86_64-linux-gnu/bits/string_fortified.h:29
#6  tracy::MemWrite<unsigned long> (val=<optimized out>, ptr=<optimized out>) at tracy/client/../common/TracyAlign.hpp:22
#7  tracy::Profiler::SendCallstack (depth=<optimized out>) at tracy/client/TracyProfiler.hpp:539
#8  ___tracy_emit_zone_begin_alloc_callstack (srcloc=140222099853920, depth=<optimized out>, active=1) at tracy/client/TracyProfiler.cpp:3483
#9  0x000055c8703456fb in tracy_client::Span::new (name=..., function=..., file=..., line=5, callstack_depth=62)
    at cargo/registry/src/github.com-1ecc6299db9ec823/tracy-client-0.12.6/src/lib.rs:92
#10 0x000055c870344f80 in repro::main::{closure#0} () at client/src/bin/repro.rs:5
#11 0x000055c870344f06 in std::sys_common::backtrace::__rust_begin_short_backtrace<repro::main::{closure#0}, ()> (f=...)
    at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/sys_common/backtrace.rs:125
#12 0x000055c8703452bb in std::thread::{impl#0}::spawn_unchecked::{closure#0}::{closure#0}<repro::main::{closure#0}, ()> ()
    at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/thread/mod.rs:481
#13 core::panic::unwind_safe::{impl#23}::call_once<(), std::thread::{impl#0}::spawn_unchecked::{closure#0}::{closure#0}> (self=..., _args=<optimized out>)
    at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/panic/unwind_safe.rs:271
#14 std::panicking::try::do_call<core::panic::unwind_safe::AssertUnwindSafe<std::thread::{impl#0}::spawn_unchecked::{closure#0}::{closure#0}>, ()> (data=<optimized out>)
    at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:403
#15 std::panicking::try<(), core::panic::unwind_safe::AssertUnwindSafe<std::thread::{impl#0}::spawn_unchecked::{closure#0}::{closure#0}>> (f=...)
    at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panicking.rs:367
#16 std::panic::catch_unwind<core::panic::unwind_safe::AssertUnwindSafe<std::thread::{impl#0}::spawn_unchecked::{closure#0}::{closure#0}>, ()> (f=...)
    at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/panic.rs:129
#17 std::thread::{impl#0}::spawn_unchecked::{closure#0}<repro::main::{closure#0}, ()> () at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/thread/mod.rs:480
#18 core::ops::function::FnOnce::call_once<std::thread::{impl#0}::spawn_unchecked::{closure#0}, ()> ()
    at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/ops/function.rs:227
#19 0x000055c8703835b3 in alloc::boxed::{impl#44}::call_once<(), dyn core::ops::function::FnOnce<(), Output=()>, alloc::alloc::Global> ()
    at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/alloc/src/boxed.rs:1636
#20 alloc::boxed::{impl#44}::call_once<(), alloc::boxed::Box<dyn core::ops::function::FnOnce<(), Output=()>, alloc::alloc::Global>, alloc::alloc::Global> ()
    at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/alloc/src/boxed.rs:1636
#21 std::sys::unix::thread::{impl#2}::new::thread_start () at library/std/src/sys/unix/thread.rs:106
#22 0x00007f880968b450 in start_thread (arg=0x7f8802edc640) at pthread_create.c:473
#23 0x00007f880945dd53 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

nagisa:
So this has a couple of major breaking changes:

* First, it is now required that users of these libraries do manually initialize the client;
  * This was necessary to fix some of the crashes that can occur when profiling. In practice I would expect that people can simply insert `tracy_client::enable` at the top of their main and be done with it.
  * This, I think, sort of breaks the no-exit feature but in return there's no longer a concern of leaking information about the binary left and right by just linking this crate in accidentally. If you don't enable the client, it should effectively behave as if it wasn't linked in the first place.
* I also went through and added all the new defines that can be toggled as features, and wired them through all the crates.

And some of the conveniences:

* `tracy_client` now re-exports `sys` and `tracing_tracy` now reexports `client`. For users who need to rely on having multiple of the crates at the same time this reduces the minutiae of having to maintain version annotations for all the crates in `Cargo.toml` (though you still can do so if you wish to retain control).

Fixes #29
Closes #32


tracy-client/src/lib.rs
```diff
@@ -1,23 +1,137 @@
...
+static ENABLE_STATE: AtomicUsize = AtomicUsize::new(0);
+const REFCOUNT_MASK: usize = usize::max_value() >> 2;
+const STATE_STEP: usize = REFCOUNT_MASK + 1; // Move forward by 1 step in the FSM
+const STATE_DISABLED: usize = 0;
+const STATE_ENABLING: usize = STATE_DISABLED + STATE_STEP;
+const STATE_ENABLED: usize = STATE_ENABLING + STATE_STEP;
+const STATE_DISABLING: usize = STATE_ENABLED + STATE_STEP;
+const STATE_MASK: usize = STATE_DISABLING;
+
+pub fn enable() {
```
nagisa, line 67: TODO: these need documentation.

I'm also torn between making the enable and disable two distinct operations vs getting some sort of guard setup going here...

tracy-client/src/lib.rs
```diff
@@ -459,37 +574,39 @@ const fn adjust_stack_depth(depth: u16) -> u16 {
 mod tests {
     use super::*;
 
-    #[global_allocator]
-    static GLOBAL: ProfiledAllocator<alloc::System> = ProfiledAllocator::new(alloc::System, 100);
```
nagisa, line 463: This in particular stopped working with the `libtest` harness because `libtest` creates some allocations before we get an opportunity to enable the client, and so some allocations get lost, and tracy is not very happy about it.

tracy-client/src/lib.rs
Ralith: A guard would make it much harder to fail to disable gracefully on exit, and is rustier.

tracy-client/src/lib.rs
Ralith: Could `ProfiledAllocator` itself call `enable` if necessary?

tracy-client/src/lib.rs
nagisa: Right, OTOH, not calling `disable` at all is going to be more correct out of the box most of the time.

As soon as the process ends up with a detached thread running instrumented code, calling `disable` would lead to a crash. And Rust code generally _loves_ having detached threads running things...

tracy-client/src/lib.rs
nagisa: Well, there are two options I think if we wanted to make this transparent to the users: go back to life-before-main and `enable` when `ProfiledAllocator` is “constructed”. Or alternatively we could call `enable`/`disable` for each allocation/deallocation, at a cost of some extra overhead for each allocation.

I think either option is pretty subpar.

tracy-client/src/lib.rs
Ralith: > And Rust code generally loves having detached threads running things...

This is certainly not true of any Rust I have ever had a hand in. Still, if failing to call `disable` is harmless, and calling it prematurely is unsound, that's a pretty good argument to leave it off the default path.

tracy-client/src/lib.rs
```diff
@@ -1,32 +1,203 @@
...
+/// because a span guard is dropped after `disable` is called.
+///
+/// ```rust,no_run
+/// {
+///     tracy_client::enable();
+///     let _span = tracy_client::span!("some span");
+///     tracy_client::disable();
+/// } // `_span` is dropped here, after the client has been disabled!
+/// ```
+pub fn disable() {
```
Ralith, line 158: From above discussion, it sounds like this must be `unsafe`. Unless calls gracefully panic when disabled?

nagisa: Right, they don’t – in practice the upstream project does not commit to any particular behaviour when the other APIs are called while the client is disabled (which is why we'd observe crashes due to things like UAF before as well.)

That I feel kind of forces our hand to heavily refactor the API of this crate, now that I think about it. So that it isn't possible to call any of the instrumentation before enabling, which I feel kind of leads us back to setting up some context type that would indicate the client is enabled and expose all the other things as methods.

Ralith: A global atomic bool which bypasses the calls when disabled would be another option which better preserves the current pattern, and I think is justifiable in this context as there's global state regardless.

tracy-client/tests/tests.rs
```diff
@@ -0,0 +1,91 @@
+use tracy_client::*;
+
+#[global_allocator]
+static GLOBAL: ProfiledAllocator<std::alloc::System> =
+    ProfiledAllocator::new(std::alloc::System, 100);
+
+fn basic_zone() {
+    let client = Client::enable();
+    let span = client.span(srcloc!("basic_zone"), 100);
```
Ralith, line 9: I think this is a big step backwards in usability. Compare: `span!("basic_zone")`.

nagisa: It sure is. It wouldn't be a huge problem to add the macro back in, although it would have to call a `Client::enable` for each invocation. I guess I could also make it optionally take the client as an argument too.

Ralith: An alternative approach would be to track enabled-ness globally with an `AtomicBool`; then nobody has to deal with `Client`s and the existing API can be preserved.

Ralith:
FWIW, this is too big a reduction in usability for me; I will get by on the previous release for as long as I can, and then fork or roll my own front-end on top of -sys when there is no other option.

nagisa:
I encourage PRs and/or design discussion. My primary use is through the tracing facade and so the primary requirement I have is for tracing-tracy to work reliably. Which has turned out to be a nontrivial problem. I'd be happy to see the client crate be more ergonomic (or recover whatever ergonomics it has lost) so long as it doesn't reintroduce the known reliability issues.

maxammann:
At maplibre-rs we will continue to use this crate :) but I also see issues with the following usage:

```
    let client = Client::start();
    let span = client.span(span_location!("basic_zone"), 100);
    span.emit_value(42);
    span.emit_text("some text");
    for i in 322..420 {
        span.emit_value(i);
    }
```

Library users will end up putting the `client` into a global variable. I think such an API should be available in the crate for simpler usage. 

I think we should keep the previous API, such that users can upgrade easily. The breaking change from 0.7 to 0.8 should be that you have to call `tracy_init()` in the beginning of the main. After adding that to the main the following still works `span!("basic_zone")`.

nagisa:
For spans, the `span!` macro [still exists](https://docs.rs/tracy-client/latest/tracy_client/macro.span.html), is [tested here](https://github.com/nagisa/rust_tracy_client/blob/60e0620c4f262a532142141ec44e37b3177cdd8c/tracy-client/tests/tests.rs#L60-L70) and is largely unchanged in its usage – the only difference from the previous variation is that it will panic if the `Client::start()` hasn't been called beforehand.

Admittedly a fair amount of the rest of the API (plots, frame marks, messages, setting a thread name…) did regress since they aren't free functions or macros anymore. I imagine adding them back wouldn't be too onerous if somebody relied on them. In fact, I suspect adding a `message!` macro that allowed `format_args!`-style formatting would be an improvement over the previous status-quo. With you lot being the actual users of the API provided by this crate I think y'all are in the best position to actually achieve something here ^^

There are of course some changes that are not addressable to the best of my knowledge, as noted in the PR description:

* The `no-exit` crate feature, as `Tracy` is no longer registering the `atexit` handlers anymore;
* The fact that we pretty much just leak a bunch of client data at process exit could make some memory analysis tools not super happy?

Ralith:
Ah, thanks, that makes the new status quo much clearer! I'll be happy to contribute convenience macros as the need arises, then.

PR #36, opened 2022-05-01
Define helper macros for trace output methods on Client

Ralith:
Not yet tested with Tracy, but these are pretty simple.

Discussion points:
- IMO these would be more convenient if they no-op'd rather than panicing when enabled at compiletime but disabled at runtime. In particular, this would make it more useful to use tracy-enabled binaries routinely, whereas IIRC enabled-but-unused tracy code is very leaky due to buffering? However, this would require adding state and branching to the `Span` and `Frame` objects.
- `frame_mark` doesn't have to be a macro, so it isn't. This might be surprising, since every other convenience interface is a macro. OTOH, most code bases will probably call this in exactly one place, so maybe no big deal?

### DIFF
--- a/tracy-client/src/frame.rs
+++ b/tracy-client/src/frame.rs
@@ -108,3 +108,42 @@ impl Drop for Frame {
         }
     }
 }
+
+/// Convenience shortcut for [`Client::frame_mark`] on the current client.
+///
+/// # Panics
+///
+/// - If a `Client` isn't currently running.
+pub fn frame_mark() {
+    Client::running()
+        .expect("frame_mark! without a running Client")
+        .frame_mark();
+}
+
+/// Convenience macro for [`Client::secondary_frame_mark`] on the current client.
+///
+/// # Panics
+///
+/// - If a `Client` isn't currently running.
+#[macro_export]
+macro_rules! secondary_frame_mark {
+    ($name: literal) => {{
+        $crate::Client::running()
+            .expect("secondary_frame_mark! without a running Client")
+            .secondary_frame_mark($crate::frame_name!($name))
+    }};
+}
+
+/// Convenience macro for [`Client::non_continuous_frame`] on the current client.
+///
+/// # Panics
+///
+/// - If a `Client` isn't currently running.
+#[macro_export]
+macro_rules! non_continuous_frame {
+    ($name: literal) => {{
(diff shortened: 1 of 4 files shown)


Ralith:
> However, this would require adding state and branching to the `Span` and `Frame` objects.

One possible mitigation would be to return `Option<_>` from the macros, though this reduces usability when `Span` methods are actually called.

Ralith:
CI failure is on the doctest at the top of tracing-tracy/src/lib.rs, and only on stable/windows/all features. Don't think it's anything this PR touches. Looks like an UB crash of some sort?

tracy-client/src/frame.rs
```diff
@@ -108,3 +108,42 @@ impl Drop for Frame {
         }
     }
 }
+
+/// Convenience macro for [`Client::frame_mark`] on the current client.
```
nagisa, line 112:
```suggestion
/// Convenience shortcut for [`Client::frame_mark`] on the running client.
```

nagisa:
> CI failure is on the doctest at the top of tracing-tracy/src/lib.rs, and only on stable/windows/all features. Don't think it's anything this PR touches. Looks like an UB crash of some sort?

This is a 0xc0000005 which is the traditional SIGSEGV we know and ~~love~~ hate. Given that the test suite of this crate is has been very prone to triggering these in the past for one reason or another, I'm not at all surprised to see these (though I did hope that this would get fixed by the use of manual lifetime management.)

> One possible mitigation would be to return Option<_> from the macros, though this reduces usability when Span methods are actually called.

Right, I pondered that idea, but figured that in the end people won't love the idea of having to do `span.map(|s| s.method())` for method spans. Though there's nothing stopping us from adding yet another macro that does not require a running client.

Ralith:
> people won't love the idea of having to do `span.map(|s| s.method())`

Another idea: `struct MaybeSpan(Option<Span>);` that just encapsulates this pattern, used only(?) for the macro. Bit of extra boilerplate to maintain, but I don't mind writing it once and I don't see the `Span` API churning all that much in the long run.

nagisa:
Thanks! We can probably hash out the span story further in a separate PR.

Ralith:
Fair point, *removing* panics is always backwards compatible.

PR #37, opened 2022-05-07
Switch literal to expr so that stringify is useable.

Imberflur:
See original reasoning here https://github.com/nagisa/rust_tracy_client/pull/23#issuecomment-950406040


Ralith (approved):
Whoops, thanks!

PR #38, opened 2022-05-23
Set Tracy thread names according to Rust std thread names

infmagic2047:
Rust's std library has the concept of [named threads](https://doc.rust-lang.org/std/thread/#naming-threads), but the names are currently not passed to Tracy.

This PR adds `set_thread_name_from_std()` in tracy-client for passing the std thread names to Tracy, and does so automatically in tracing-tracy when the first span or event is handled for each thread.

### DIFF
--- a/tracy-client/src/lib.rs
+++ b/tracy-client/src/lib.rs
@@ -148,6 +148,23 @@ impl Client {
             internal::set_thread_name(name.as_ptr().cast());
         }
     }
+
+    /// Set the current thread name according to Rust's [`Thread::name()`].
+    ///
+    /// This is useful if the thread name is already set with Rust's standard library APIs, and you
+    /// want to copy the name to Tracy. Does nothing if the current thread is not named according
+    /// to [`Thread::name()`].
+    ///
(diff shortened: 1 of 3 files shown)


tracy-client/src/lib.rs
```diff
@@ -148,6 +148,23 @@ impl Client {
...
+    /// [`Thread::name()`]: std::thread::Thread::name
+    pub fn set_thread_name_from_std(&self) {
+        #[cfg(feature = "enable")]
+        {
+            if let Some(name) = std::thread::current().name() {
+                // This will not panic because Rust disallows thread names with null bytes.
+                self.set_thread_name(name);
+            }
+        }
+    }
```
Ralith, line 167: This is trivial for downstream code to write themselves; is it really worth having a method for it?

infmagic2047: I put the method here to raise some awareness of the Rust thread naming API, as `set_thread_name` alone may imply thread naming being some Tracy-specific feature.

Ralith: Could that be solved by pure documentation instead?

PR #40, opened 2022-05-25
Avoid removing pub from members named type

### LINKED ISSUE #39: __tracy_gpu_new_context_data has private field making it unconstructable
Oddly it seems to have a completely private field which we need to fill to be able to call the function. I'm not sure what caused bindgen to emit this one field as private.

https://docs.rs/tracy-client-sys/latest/tracy_client_sys/struct.___tracy_gpu_new_context_data.html

cwfitzgerald:
Closes #39

Not sure if there's anything else I need to do to make the regeneration happen right.

### DIFF
--- a/make_sys.sh
+++ b/make_sys.sh
@@ -34,7 +34,8 @@ bindgen "$BASEDIR/TracyC.h" \
   --disable-header-comment \
   -- \
   ${DEFINES[@]}
-sed -i 's/pub type/type/g' 'tracy-client-sys/src/generated.rs'
+# The space after type avoids hitting members called "type".
+sed -i 's/pub type /type/g' 'tracy-client-sys/src/generated.rs'
 
 for REQUIRED_FILE in ${REQUIRED[@]}
 do


cwfitzgerald:
Alright, I'll just wait for CI to regen then, thanks!

nagisa:
Merged, thanks!

Imberflur:
```sh
's/pub type /type/g'
                ^
```
this probably needs a space here to avoid removing the space

nagisa:
Yeah, I have added that in a follow-up [here](https://github.com/nagisa/rust_tracy_client/commit/54ebfa92a59928eefae5030a76588312300290b0).

Ralith:
A more general/robust solution than sed might be to pass allowlist regexps to bindgen.

nagisa:
I remember trying to do that back when this was initially added. I don't remember exactly why it didn't work but it must have been either one of:

1. It doesn't work for system header stuff that the API depends on and bindgen still publicly exports the types;
2. Bindings no longer compile due to missing type definitions.

PR #41, opened 2022-05-25
Add Safe GPU Api

cwfitzgerald:
Depends on #39 

Context https://github.com/Wumpf/wgpu-profiler/issues/15

This adds the GPU api to tracy_client. It merely adds safe interfaces. I am putting this up as a draft to get opinions on architecture.

TODO:
- [x] docs
- [x] doctests


tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[cfg(feature = "enable")]
+    context: u8,
+}
+#[cfg(feature = "enable")]
+static GPU_CONTEXT_INDEX: AtomicU8 = AtomicU8::new(0);
+
+///
+pub struct GpuSpan {
+    #[cfg(feature = "enable")]
+    _client: Client,
```
Ralith, line 45: `Client` is already a ZST, might not be worth bothering with feature-gating it.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[cfg(feature = "enable")]
+    end_query_id: Option<u16>,
+}
+#[cfg(feature = "enable")]
+static GPU_SPAN_INDEX: AtomicU16 = AtomicU16::new(0);
+
+impl Client {
+    ///
+    pub fn new_gpu_context(
+        self,
```
Ralith, line 59: Should this be `&self`?

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+                    context,
+                    flags: 0,
+                    type_: ty as u8,
+                },
+            ));
+
+            if let Some(name) = name {
+                sys::___tracy_emit_gpu_context_name_serial(sys::___tracy_gpu_context_name_data {
+                    context,
+                    name: name.as_ptr() as *const i8,
```
Ralith, line 82: nit:
```suggestion
                    name: name.as_ptr().cast(),
```

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+        GpuContext {}
+    }
+}
+
+impl GpuContext {
+    ///
+    pub fn span_alloc(&self, name: &str, function: &str, file: &str, line: u32) -> GpuSpan {
+        #[cfg(feature = "enable")]
+        unsafe {
+            let srcloc = sys::___tracy_alloc_srcloc_name(
```
Ralith, line 102: We should expose a non-allocating interface as well; see `Client::span` for an example.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+            }
+        }
+        #[cfg(not(feature = "enable"))]
+        GpuSpan {}
+    }
+}
+
+impl GpuSpan {
+    ///
+    pub fn end_zone(&mut self) {
```
Ralith, line 134: Is it sensible to call this multiple times?

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+            sys::___tracy_emit_gpu_zone_end_serial(sys::___tracy_gpu_zone_end_data {
+                queryId: query_id,
+                context: self.context,
+            });
+            self.end_query_id = Some(query_id);
+        }
+    }
+
+    ///
+    pub fn upload_timestamp(self, start_timestamp: i64, end_timestamp: i64) {
```
Ralith, line 147: Where do these timestamps come from?

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+        unsafe {
+            let context = GPU_CONTEXT_INDEX.fetch_add(1, Ordering::Relaxed);
+
+            sys::___tracy_emit_gpu_new_context(std::mem::transmute(
+                ___tracy_gpu_new_context_data {
+                    gpu_time: gpu_timestamp,
+                    period,
+                    context,
+                    flags: 0,
+                    type_: ty as u8,
```
Ralith, line 75: If we're relying on particular values for this enum, I'd prefer to declare them explicitly in the enum definition.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub fn new_gpu_context(
+        self,
+        name: Option<&str>,
+        ty: GpuContextType,
+        gpu_timestamp: i64,
+        period: f32,
+    ) -> GpuContext {
+        #[cfg(feature = "enable")]
+        unsafe {
+            let context = GPU_CONTEXT_INDEX.fetch_add(1, Ordering::Relaxed);
```
Ralith, line 67: What happens when this overflows?

Ralith (changes requested):
I'd like to see this expanded on with concise, ergonomic, non-allocating macros as has been done with e.g. `span`.

I'm also a bit confused about how this works. There doesn't seem to be much here that could plausibly interact with the GPU.

tracy-client/src/gpu.rs
cwfitzgerald: There is no non-allocating interface for gpu timestamps.

tracy-client/src/gpu.rs
cwfitzgerald: No.

tracy-client/src/gpu.rs
cwfitzgerald: I'm just following the pattern used by other code, client methods are only really used by doing `Client::running().unwrap().whatever()` and consuming the client.

tracy-client/src/gpu.rs
cwfitzgerald: The corresponding queries that corresponded with the calls to start and end scope.

tracy-client/src/gpu.rs
cwfitzgerald: Dunno, tracy's documentation doesn't say what happens if you try to create a second version of the same context number.

tracy-client/src/gpu.rs
Ralith: Then the interface should prevent the user from doing so.

tracy-client/src/gpu.rs
Ralith: Then we should not presume it is sound to do so, and a safe interface must therefore prevent it, e.g. by panicing or nooping.

tracy-client/src/gpu.rs
Ralith: Hmm, I wonder if that pattern was really intended. It's pretty weird to not be able to reuse a `Client` reference.

tracy-client/src/gpu.rs
Ralith: To be specific, we should expose an interface which does not call `___tracy_alloc_srcloc_name`. While `___tracy_emit_gpu_zone_begin_alloc_serial` *may* allocate, it uses shared storage internally which is exponentially resized, making it considerably cheaper on average.

tracy-client/src/gpu.rs
Ralith: Does tracy leave it entirely up to the user to issue those queries? What's the point of passing in a `GpuContextType`, then?

tracy-client/src/gpu.rs
Ralith: On closer examination, it's probably easier to feature-gate than not, so nevermind.

tracy-client/src/gpu.rs
cwfitzgerald: I need to pull a srcloc from somewhere though right? There's no gpu api that takes source information without allocating.

Ralith: Oh, I see; that's disappointing. Would be nice to open an issue upstream, but not a blocker here, then.

tracy-client/src/gpu.rs
cwfitzgerald: yes.

Literally just to decide which pretty label to display here

image

Ralith: Well, can't fault them for keeping the interface minimal; I guess it's on the community to provide bindings to whatever frontends.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+                name.len(),
+            );
+
+            let query_id = GPU_SPAN_INDEX.fetch_add(1, Ordering::Relaxed);
+
+            sys::___tracy_emit_gpu_zone_begin_alloc_serial(sys::___tracy_gpu_zone_begin_data {
+                srcloc,
+                queryId: query_id,
+                context: self.context,
+            });
```
nagisa, line 118: The handling of `queryId` here seems not quite right. First, AFAICT these are supposed to be tracked per-context, rather than globally. And second, Tracy implementations appear to be wrapping this number modulo some sort of a array/pool size. This is important for collecting the timestamps as far as I can tell. My understanding is that these IDs should also be made accessible to the holder of the `GpuSpan` somehow so that they could implement their own timestamp queries.

nagisa (commented):
With the caveat that I don't deal with GPUs much, the overall shape seems quite reasonable to me. Excited to see progress here, and sorry for the delay in reviewing this.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+                    gpu_time: gpu_timestamp,
+                    period,
+                    context,
+                    flags: 0,
+                    type_: ty as u8,
+                },
+            ));
+
+            if let Some(name) = name {
+                sys::___tracy_emit_gpu_context_name_serial(sys::___tracy_gpu_context_name_data {
```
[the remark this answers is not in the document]
cwfitzgerald: Why couldn't this be shared? I've designed it so it should be able to.

tracy-client/src/gpu.rs
cwfitzgerald: Tracking them per context would be fine - I'll just move to a local atomic.

Why would the implementer need to know the IDs at all? Once they make a gpu span, they create a timestamp query for the beginning of the span, then when they end it they register a timestamp query for the end. Once they both resolve, they call this function to upload the values for those timestamps.

cwfitzgerald:
Pardon the delay, most of these should be addressed now.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,171 @@
...
+        #[cfg(feature = "enable")]
+        unsafe {
+            // We use a mutex to lock the context index to prevent races when using fetch_add.
+            let mut context_index_guard = GPU_CONTEXT_INDEX.lock().unwrap();
+            if *context_index_guard == 255 {
+                return None;
+            }
+            let context = *context_index_guard;
+            *context_index_guard += 1;
+            drop(context_index_guard);
```
nagisa, line 70: This could still use atomics to a great effect, but I’m okay with keeping this as is for the time being too.

tracy-client/src/gpu.rs
nagisa: Still needs to be done.

tracy-client/src/gpu.rs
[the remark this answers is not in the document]
nagisa: Well it wasn’t clone before IIRC.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }
+        #[cfg(not(feature = "enable"))]
+        GpuSpan {}
+    }
+}
+
+impl GpuSpan {
+    ///
+    pub fn end_zone(&mut self) -> bool {
```
nagisa, line 139: Hm, how is this method supposed to be used? Can you please add a test/example here? It’ll make it easier to discuss the API here.

nagisa (commented):
Overall this looks good enough to me, just a couple of nits (and docs and tests) before this could be merged.

Sorry for the long delays in review here. Holidays and stuff.

tracy-client/src/gpu.rs
cwfitzgerald: I don't think we can - you could end up causing UB if your thread timing is really unfortunate.

If the index is 255.

T0 calls atomic_add(1). Before we get to the conditional we're suspended. T1 comes along and tries to increment it, sees that it's value is 0, goes ahead and stomps on the old (still outstanding) gpu context, causing all sorts of nonsense. I think we need to guarantee mutual exclusion when the check is happening.

nagisa: ah `atomic_add` was not what I was thinking of. `compare_exchange` could be used to implement a saturating u8 add, so value could be made to never wrap or exceed `255`.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,163 @@
...
+    gpu_time: i64,
+    period: f32,
+    context: u8,
+    flags: u8,
+    type_: u8,
+}
+
+#[repr(u8)]
+///
+pub enum GpuContextType {
```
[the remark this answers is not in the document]
cwfitzgerald: Unfortunately I don't see an easy way to, considering it's only defined in the TracyQueue.hpp file

tracy-client/src/gpu.rs
cwfitzgerald: Explicit values specified.

cwfitzgerald:
This totally got lost in my history - got this fully back up to date, an improved some of the code to make it more resilient.

Needs more documentation and the test causes tracy to crash, which I need to get a debug build of tracy to debug (this was an issue before the fix, the test just makes it trivial to reproduce.

tracy-client/src/gpu.rs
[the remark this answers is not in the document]
nagisa: Ah, I think I meant that that this would be a contribution to the tracy project to add the enum definition to the `TracyC.h` file. In my past experience the author of the project is receptive to changes like these.

(Of course I wouldn’t want to block landing of this PR on a new release of Tracy, so it is fine to have this here as-is for the time being)

nagisa:
I would be happy to land this as soon as the doc-comments are filled in.

Tracy crashes: unfortunately those *usually* are an issue in the upstream project. I’m of an opinion that we’re not supposed to work around those in the bindings. The only place where any meaningful amount of complex logic is warranted is in the `tracing` bridge which attempts to paper over the semantic mismatches between two ecosystems. When reporting an issue upstream, do beware that the author there does not have particularly warm feelings towards the Rust ecosystem, so I found it better/easier to come up with minimal C code samples, or to straight=up provide fixes.

cwfitzgerald:
Cool! Will add the doc comments and such and some end-to-end documentation.

I want to at least get a culpret for the tracy crash, to make sure we're not somehow misusing the api, as the builtin integrations seem to work okay. Will build a c repro if it happens, thankfully should be pretty simple.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,201 @@
...
+                span_index: Arc::new(AtomicU16::new(0)),
+            })
+        }
+        #[cfg(not(feature = "enable"))]
+        Some(GpuContext {})
+    }
+}
+
+impl GpuContext {
+    ///
```
Ralith, line 119: Missing doc comments here and below?

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,201 @@
...
+            }
+        }
+        #[cfg(not(feature = "enable"))]
+        GpuSpan {}
+    }
+}
+
+impl GpuSpan {
+    ///
+    pub fn end_zone(&mut self) -> Option<()> {
```
Ralith, line 160: What's the significance of this call is if we're passing timestamps explicitly anyway? Is tracy overloading this API to track the CPU time involved in recording commands as well?

(should probably be answered in the doc comment, not just the PR)

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,201 @@
...
+                    file.len(),
+                    function.as_ptr().cast(),
+                    function.len(),
+                    name.as_ptr().cast(),
+                    name.len(),
+                )
+            };
+
+            // Allocate two ids, one for the start value, one for the end.
+            let query_id = self.span_index.fetch_add(2, Ordering::Relaxed);
```
Ralith, line 136: When are these IDs freed?

Ralith (changes requested):

tracy-client/src/gpu.rs
cwfitzgerald: Tracy correlates the gpu side timestamps with cpu side timestamps - each creation call is where tracy pulls the cpu side timestamps from.

tracy-client/src/gpu.rs
cwfitzgerald: My understanding is once the spans ids have timestamps associated with them they can be reused again. The consequences of reusing queries are not exactly clear to me (this api has basically no documentation), and we might need to add a freelist to ensure the is no improper reuse.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,201 @@
...
+        Some(())
+    }
+
+    ///
+    pub fn upload_timestamp(self, start_timestamp: i64, end_timestamp: i64) {
+        #[cfg(feature = "enable")]
+        {
+            if !self.ended {
+                return;
+            }
```
Ralith, line 183: Should this be an `assert`? Silently ignoring timestamps if someone forgets to call `end_zone` might be frustrating.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,201 @@
...
+            unsafe {
+                sys::___tracy_emit_gpu_time_serial(sys::___tracy_gpu_time_data {
```
Ralith, line 185: Is it safe to call this twice with the same ID without reallocating it?

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+impl GpuSpan {
+    ///
+    pub fn end_zone(&mut self) -> Option<()> {
+        #[cfg(feature = "enable")]
+        {
+            if self.ended {
+                return None;
```
Ralith, line 164: What's the user intended to do with this return value? Should we use a `Result` with a sensible (ZST) error so `unwrap` gives intelligible feedback? Alternatively, would it be that bad to silently eat redundant calls?

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,201 @@
...
+    #[cfg(feature = "enable")]
+    context: u8,
+    #[cfg(feature = "enable")]
+    span_index: Arc<AtomicU16>,
+}
+#[cfg(feature = "enable")]
+static GPU_CONTEXT_INDEX: Lazy<Mutex<u8>> = Lazy::new(|| Mutex::new(0));
+
+///
+pub struct GpuSpan {
```
Ralith, line 48: Maybe mark this `#[must_use]`?

tracy-client/src/gpu.rs
Ralith: At a guess, perhaps `___tracy_emit_gpu_time_serial` is an implicit dealloc? In that case we'd need to do something in `Drop`.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,201 @@
...
+            })
+        }
+        #[cfg(not(feature = "enable"))]
+        Some(GpuContext {})
+    }
+}
+
+impl GpuContext {
+    ///
+    pub fn span_alloc(&self, name: &str, function: &str, file: &str, line: u32) -> GpuSpan {
```
Ralith, line 120: Could we also have a non-`_alloc` variant that takes `&'static SpanLocation`?

cwfitzgerald: Yes! Such an api didn't exist in the c api when I first wrote this MR, but it does now, so I'll add it.

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,201 @@
...
+    OpenCL = 3,
+    /// A D3D12 context.
+    Direct3D12 = 4,
+    /// A D3D11 context.
+    Direct3D11 = 5,
+}
+
+///
+#[derive(Clone)]
+pub struct GpuContext {
```
Ralith, line 36: When the `"enable"` feature is unset, this struct's definition will become public. Include a private field to mitigate this. Similar for `GpuSpan`.

Ralith (changes requested):

cwfitzgerald:
Traced down the bug, it was actually an assert in tracy that was getting hit. I wasn't using the `_serial` version of the context naming function, which caused a failure.

From looking at the tracy's use of query ids, they have asserts that they are never re-used until the timestamps are written. As such, I'm going to add a freelist for validation and re-use prevention.

tracy-client/src/gpu.rs
cwfitzgerald: It would be a little weird, but would probably be okay to eat double calls.

tracy-client/src/gpu.rs
cwfitzgerald: I'm not sure what we'd do in a drop. Maybe write out junk zero-value timestamps? Or write out the same timestamp as the start of the gpu context, so they don't interfere with other spans?

tracy-client/src/gpu.rs
cwfitzgerald: Resolved to eat the double calls.

tracy-client/src/gpu.rs
cwfitzgerald: Changed to an assert

tracy-client/src/gpu.rs
Ralith: The second sounds like the safest bet to me, seeing as Tracy's C++ API tries to make this case impossible.

tracy-client/src/gpu.rs
cwfitzgerald: Welp. Turns out the function with the different name has exactly the same argument struct as the other one, so there's no difference in the calling code :|

tracy-client/src/gpu.rs
cwfitzgerald: Done.

cwfitzgerald:
Failures unrelated.

tracy-client/src/gpu.rs
Ralith: Based on the implementation of the analogous non-GPU API, I think the intention is that we reinterpret_cast the pointer into the same field:

https://github.com/wolfpld/tracy/blob/0e26b9e53fc88aeed75050c959af751e8c48828b/public/client/TracyProfiler.cpp#L3958C61-L3958C67

I didn't dig all the way down into the guts to see where this gets decoded, but there's a tag that tracy does indeed branch on to determine how to interpret the payload data.

cwfitzgerald: Nice hint! It works!

cwfitzgerald:
I think this is good for final reviews

tracy-client/src/gpu.rs
```diff
@@ -0,0 +1,415 @@
...
+    #[cfg(feature = "enable")]
+    _client: Client,
+    #[cfg(feature = "enable")]
+    value: u8,
+    #[cfg(feature = "enable")]
+    gpu_start_timestamp: i64,
+    #[cfg(feature = "enable")]
+    span_freelist: Arc<Mutex<Vec<u16>>>,
```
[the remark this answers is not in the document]
cwfitzgerald: Yes, they make no attempt to globally de-duplicate them in their own bindings.

nagisa:
Merged this in https://github.com/nagisa/rust_tracy_client/commit/9ed943bd6b9185d824bf79906c9775eda4edb70e and pushed out a new release.

cwfitzgerald:
The commit on main unfortunately appears to be a fairly old iteration of this PR, missing most of the doc comments and some of the fixes.

nagisa:
Ah, my bad, looks like squash-merging this on a late sunday night was a mistake :)

Yanked 0.16.0, merged the rest of the PR, and released a new one. I believe it correctly contains the rest of this PR.

PR #43, opened 2022-06-11
Set default callstack_depth to 0

nagisa:
Collecting callstacks is quite expensive and can make each
instrumentation point ~100 times slower (200ns -> 20µs). Originally
callstack collection was enabled by default in a hope to provide a great
out-of-the-box experience by just enabling the layer or inserting some
instrumentation points. In retrospective that is a poor default.

Compared to figuring out why the performance is suffering, it seems like
it would be _much_ easier for user to discover how to enable collection
of stack frames when they find themselves wanting to see some in their
profiles. It also gives us an opportunity to visibly document the
pitfalls of enabling collection of the stack frames.

### DIFF
--- a/tracing-tracy/src/lib.rs
+++ b/tracing-tracy/src/lib.rs
@@ -85,7 +85,7 @@ impl TracyLayer<DefaultFields> {
     pub fn new() -> Self {
         Self {
             fmt: DefaultFields::default(),
-            stack_depth: 64,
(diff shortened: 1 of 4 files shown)


Ralith (approved):
I mostly use explicit spans and have never actually wanted stack frame capture, so for my use case this is a substantial free optimization.

james7132 (approved):

PR #45, opened 2022-06-12
Cache buffers, allow omitting span names, options for not formatting KVs all the time

nagisa:
With this plain `error!("banana")` can take as little as 200ns (~0 overhead) but it will also lead to no meaningful messages if you have `error!(someotherkey)`, and it will also drop all the KVs too.

Similarly, emitting a span can happen in “just” 1.2µs which is also pretty nice.

Finally, `tracy_client::span!()` “just works” and produces meaningful trace outputs, which is _very_ nice. I think this one is likely uncontroversial and can be landed as is (see 2nd commit).

Otherwise this is just a code/brain dump to which I don't anticipate returning for a foreseeable future. If somebody/anybody wants to finish this up somehow, I encourage you to do so ^^.


nagisa:
I went ahead and merged the 2nd commit, since improving on it is difficult, others though…

PR #49, opened 2022-07-31
Support linking against prebuilt tracy client

guusw:
I'm using this crate in a scenario with both rust and C++ code using the same tracy client, I'd like to be able to manually specify the tracy client to link against instead of having it hard-coded inside this crate, which would cause duplicate defined symbols.

Here I propose to add the following environment variables:
- TRACY_CLIENT_LIB: when set, links against this tracy client library instead of building it
- TRACY_CLIENT_LIB_PATH: when TRACY_CLIENT_LIB is set, this may contain additional library search paths

The approach is similar to https://crates.io/crates/openssl-sys , which also has a set of environment variables that control how the C/C++ code is linked/compiled

### DIFF
--- a/tracy-client-sys/build.rs
+++ b/tracy-client-sys/build.rs
@@ -50,7 +50,7 @@ fn set_feature_defines(mut c: cc::Build) -> cc::Build {
 fn main() {
     if let Ok(lib) = std::env::var("TRACY_CLIENT_LIB") {
         if let Ok(lib_path) = std::env::var("TRACY_CLIENT_LIB_PATH") {
-            println!("cargo:rustc-link-search={}", lib_path);
+            println!("cargo:rustc-link-search=native={}", lib_path);
         }
         println!("cargo:rustc-link-lib={}", lib);
     } else {
--- a/tracy-client-sys/build.rs
+++ b/tracy-client-sys/build.rs
@@ -48,25 +48,32 @@ fn set_feature_defines(mut c: cc::Build) -> cc::Build {
 }
 
 fn main() {
-    if std::env::var_os("CARGO_FEATURE_ENABLE").is_some() {
-        set_feature_defines(cc::Build::new())
-            .file("tracy/TracyClient.cpp")
-            .define("TRACY_MANUAL_LIFETIME", None)
-            .define("TRACY_DELAYED_INIT", None)
-            .warnings(false)
-            .cpp(true)
-            .flag_if_supported("-std=c++11")
-            .compile("libtracy-client.a");
-    }
-
-    match std::env::var("CARGO_CFG_TARGET_OS") {
-        Ok(target_os) => {
-            link_libraries(&target_os);
+    if let Ok(lib) = std::env::var("TRACY_CLIENT_LIB") {
+        if let Ok(lib_path) = std::env::var("TRACY_CLIENT_LIB_PATH") {
+            println!("cargo:rustc-link-search={}", lib_path);
         }
-        Err(e) => {
-            writeln!(::std::io::stderr(), "Unable to get target_os=`{}`!", e)
-                .expect("could not report the error");
-            ::std::process::exit(0xfd);
+        println!("cargo:rustc-link-lib={}", lib);
+    } else {
+        if std::env::var_os("CARGO_FEATURE_ENABLE").is_some() {
+            set_feature_defines(cc::Build::new())
+                .file("tracy/TracyClient.cpp")
+                .define("TRACY_MANUAL_LIFETIME", None)
+                .define("TRACY_DELAYED_INIT", None)
+                .warnings(false)
+                .cpp(true)
+                .flag_if_supported("-std=c++11")
+                .compile("libtracy-client.a");
+        }
+
+        match std::env::var("CARGO_CFG_TARGET_OS") {
+            Ok(target_os) => {
+                link_libraries(&target_os);
+            }
+            Err(e) => {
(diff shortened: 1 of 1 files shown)


tracy-client-sys/build.rs
```diff
@@ -48,25 +48,32 @@ fn set_feature_defines(mut c: cc::Build) -> cc::Build {
...
-            .flag_if_supported("-std=c++11")
-            .compile("libtracy-client.a");
-    }
-
-    match std::env::var("CARGO_CFG_TARGET_OS") {
-        Ok(target_os) => {
-            link_libraries(&target_os);
+    if let Ok(lib) = std::env::var("TRACY_CLIENT_LIB") {
+        if let Ok(lib_path) = std::env::var("TRACY_CLIENT_LIB_PATH") {
+            println!("cargo:rustc-link-search={}", lib_path);
```
nagisa, line 53:
```suggestion
            println!("cargo:rustc-link-search=native={}", lib_path);
```

tracy-client-sys/build.rs
```diff
@@ -48,25 +48,32 @@ fn set_feature_defines(mut c: cc::Build) -> cc::Build {
...
         }
-        Err(e) => {
-            writeln!(::std::io::stderr(), "Unable to get target_os=`{}`!", e)
-                .expect("could not report the error");
-            ::std::process::exit(0xfd);
+        println!("cargo:rustc-link-lib={}", lib);
```
nagisa, line 55: Needs an ability to specify static vs dynamic linking, and should explicitly specify the mode here. I don’t if it makes a ton of sense to statically bundle-link the client libraries, so linking to dynamic libraries should probably be a default.

tracy-client-sys/build.rs
```diff
@@ -48,25 +48,32 @@ fn set_feature_defines(mut c: cc::Build) -> cc::Build {
...
-            .warnings(false)
-            .cpp(true)
-            .flag_if_supported("-std=c++11")
-            .compile("libtracy-client.a");
-    }
-
-    match std::env::var("CARGO_CFG_TARGET_OS") {
-        Ok(target_os) => {
-            link_libraries(&target_os);
+    if let Ok(lib) = std::env::var("TRACY_CLIENT_LIB") {
```
nagisa, line 51: Hm, while being able to specify this is nice, I wonder if we can do something to not require specifying `tracy-release` or something here if people just run `make` and don’t rename the client libraries?

I’d argue that we should at the very least have an enabled-by-default crate feature for use of `vendor`ed version of tracy.

tracy-client-sys/build.rs
```diff
@@ -48,25 +48,32 @@ fn set_feature_defines(mut c: cc::Build) -> cc::Build {
...
+        match std::env::var("CARGO_CFG_TARGET_OS") {
+            Ok(target_os) => {
+                link_libraries(&target_os);
+            }
+            Err(e) => {
+                writeln!(::std::io::stderr(), "Unable to get target_os=`{}`!", e)
+                    .expect("could not report the error");
+                ::std::process::exit(0xfd);
+            }
         }
```
nagisa, line 77: Perfect opportunity to outline into a function.

tracy-client-sys/build.rs
Ralith: > Needs an ability to specify static vs dynamic linking

Normally this problem is solved by providing only the desired form of library in the search path.

nagisa: Unfortunately traditional unixes typically just chuck all types and kinds of libraries into the same directory (`/usr/lib` or some triple-specific directory underneath).

Ralith: If you want a different form than your OS packages, you can place it elsewhere and supply the path explicitly, which should give it precedence. I'm not sure I've ever seen an environment where it's idiomatic for packages to provide both static and dynamic versions of everything, and even then you could still use custom builds.

guusw: Makes sense for non-windows platforms since they have different extensions (.so/.a).
I can add a `TRACY_CLIENT_STATIC` to switch from dynamic (default) to static, again following openssl-sys a bit since they have the same

tracy-client-sys/build.rs
guusw: I'm not sure what you mean with `tracy-release` but my intention was to not change the default behaviour so only if you specify `TRACY_CLIENT_LIB` the bundled version is not compiled.

nagisa: Well, if you want to use `TRACY_CLIENT_LIB` you must now specify the library name (the `{name}` part of `lib{name}.{so,dylib,a}` on unixes), but the tracy’s makefiles produce a client library with a well-known name. 

For cmake build that seems to be `libTracyClient.a` and the makefiles appear to spit out a `libtracy-release.so`/`libtracy-debug.so`. I guess its okay to keep this as is in that case…

tracy-client-sys/build.rs
nagisa: I’d be happy with `TRACY_CLIENT_STATIC` or whatever if it means rustc is told explicitly which one to link.

nagisa:
I went ahead and merged this alongside the `_STATIC` additions I had requested in one of the review commends in ca281d6eab216c3db7bcf96745421c8edda14a51.

Thanks for the contribution and sorry it took so long to get this landed!

guusw:
Ah my bad, I forgot about the PR. Thanks for finishing it.

PR #53, opened 2022-10-27
Update to Tracy v0.9

### LINKED ISSUE #52: Update to the new version of Tracy
The auto-update PR script broke, at least because installing bindgen split out the CLI into a separate crate. But also, Tracy itself has significant changes, so I don’t really anticipate things "just" working after some version bumps.

nagisa:
Fixes #52


nagisa:
Uh, unfortunately I cannot reproduce the CI failures locally on my W10 VM… :/

nagisa:
This is not reproducible on a fresh win2k22 install either… Very curious this only occurs on nightly, so would be very nice if this was actually reproducible, as right now this looks like potentially something genuinely serious, but I’ve no means whatsoever to actually do anything about it.

TBH I’m inlined to ignore the failure for now and just merge.

tracy-client-sys/tracy/tracy/TracyC.h
```diff
@@ -0,0 +1,341 @@
...
+TRACY_API void ___tracy_emit_gpu_zone_end( const struct ___tracy_gpu_zone_end_data data );
+TRACY_API void ___tracy_emit_gpu_time( const struct ___tracy_gpu_time_data );
+TRACY_API void ___tracy_emit_gpu_new_context( const struct ___tracy_gpu_new_context_data );
+TRACY_API void ___tracy_emit_gpu_context_name( const struct ___tracy_gpu_context_name_data );
+TRACY_API void ___tracy_emit_gpu_calibration( const struct ___tracy_gpu_calibration_data );
+
+TRACY_API void ___tracy_emit_gpu_zone_begin_serial( const struct ___tracy_gpu_zone_begin_data );
+TRACY_API void ___tracy_emit_gpu_zone_begin_callstack_serial( const struct ___tracy_gpu_zone_begin_callstack_data );
+TRACY_API void ___tracy_emit_gpu_zone_begin_alloc_serial( const struct ___tracy_gpu_zone_begin_data );
+TRACY_API void ___tracy_emit_gpu_zone_begin_alloc_callstack_serial( const struct ___tracy_gpu_zone_begin_callstack_data );
```
farnoy, line 188: These are new C APIs in Tracy 0.9, but the `sys` crate does not expose them yet. Could we add them add the same time?

make_sys.sh
```diff
@@ -18,15 +18,13 @@ DEFINES=("-DTRACY_ENABLE" "-DTRACY_MANUAL_LIFETIME" "-DTRACY_FIBERS")
...
-
 BASEDIR=("$DESTINATION"/*)
-REQUIRED=($(gcc --dependencies ${DEFINES[@]} "$BASEDIR/TracyClient.cpp" | grep -o "$BASEDIR/[^ \\]*"))
 
-mkdir -p "tracy-client-sys/tracy"
+cp -r "$BASEDIR/public/"* "tracy-client-sys/tracy/"
+cp "$BASEDIR/LICENSE" "tracy-client-sys/tracy/"
 
-bindgen "$BASEDIR/TracyC.h" \
+bindgen "tracy-client-sys/tracy/TracyC.h" \
```
farnoy, line 27: I think the issue is somewhere here, there are two `TracyC.h` files in the tree, only one of them has been updated and bindgen uses the old one, that's my theory anyway.

nagisa: Great catch. I did notice the double file while investigating the SIGSEGV and made a note in my head to investigate it, only to promptly forget all about it. Thanks for writing it down!

nagisa:
I ran out of steam to investigate the sigsegv past what I wrote down in https://github.com/wolfpld/tracy/issues/483. Either way this seems like not-a-bindings issue (except maybe for the fact that we’re using the manual lifetime mode).

PR #57, opened 2023-01-18
doc: maintain full documentation of the features in all crates

### LINKED ISSUE #46: Documentation on the available feature flags is not-trivial to discover
I have a remote server I'd like to profile more than once, however if I disconnect, it says:

```
The client you are trying to connect to is no longer able to send 
profiling data, because another server was already connected to it.
You can do the following:

1. Restart the client application.
2. Rebuild the client application with on-demand mode enabled.
```

Is there a way to enable on-demand mode?

nagisa:
People have been complaining that finding feature documentation is difficult.

Fixes #46. Hopefully.


Jerrody:
> People have been complaining that finding feature documentation is difficult.
> 
> Fixes #46. Hopefully.

Thank you！

PR #61, opened 2023-03-07
Fix version table typo

Imberflur:
I assume this was generated by `make_sys.sh` before the value for next client version was bumped from 14 to 15 here: 

https://github.com/nagisa/rust_tracy_client/blob/0763d2d16c75af71cc1c97c26027aa4cf3f7ccc9/make_sys.sh#L74


nagisa:
Thanks!

PR #71, opened 2023-09-20
Add {Frame,Plot}Name::new_leak()

### LINKED ISSUE #69: I had to use internal for dynamic FrameName
Hi, I'm using tracy-client to profile a Wayland compositor. I'm using non-continuous frames for profiling monitor redrawing. However, with multiple monitors plugged in, every monitor should have its own frame set, as they are disjoint and redraw on their own vsync cycles.

Since monitors can be plugged in and out at runtime, I have to generate FrameNames dynamically, based on the name of the connector. I understand it's not really supported by Tracy at the moment, but it works well enough™ if I leak the name string, so that it's available for the remaining program lifetime. Unfortunately, I can't use the frame_name! macro as it only accepts a literal.

```rust
let vblank_frame_name = unsafe {
    tracy_client::internal::create_frame_name(format!("vblank on {output_name}\0").leak())
};
```

(I do this once when a monitor is connected, not every frame.)

Would it make sense to expose something for this in tracy-client?

YaLTeR:
Closes #69.

### DIFF
--- a/tracy-client/src/frame.rs
+++ b/tracy-client/src/frame.rs
@@ -11,6 +11,34 @@ pub struct Frame(Client, FrameName);
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
 pub struct FrameName(pub(crate) &'static str);
 
+impl FrameName {
+    /// Construct a `FrameName` dynamically, leaking the provided String.
+    ///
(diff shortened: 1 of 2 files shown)


YaLTeR:
Thanks for merging! I see you bumped the crate version in the repo but it doesn't seem like it had made it to crates.io yet?

YaLTeR:
Thanks!

PR #85, opened 2023-12-25
Code cleanup and maintenance

SUPERCILEX:


tracing-tracy/src/lib.rs
```diff
@@ -148,64 +150,93 @@ where
...
 {
     fn on_new_span(&self, attrs: &Attributes<'_>, id: &Id, ctx: Context<'_, S>) {
-        if let Some(span) = ctx.span(id) {
-            let mut extensions = span.extensions_mut();
-            if extensions.get_mut::<FormattedFields<F>>().is_none() {
-                let mut fields = FormattedFields::<F>::new(String::with_capacity(64));
-                if self.fmt.format_fields(fields.as_writer(), attrs).is_ok() {
-                    extensions.insert(fields);
-                }
+        let Some(span) = ctx.span(id) else {
```
SUPERCILEX, line 153: This and below is just re-ordering the methods to match the trait definition.

tracing-tracy/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "tracing-tracy"
 version = "0.10.4"
 authors = ["Simonas Kazlauskas <tracing-tracy@kazlauskas.me>"]
 license = "MIT/Apache-2.0"
-edition = "2018"
+edition = "2021"
```
nagisa, line 6: Is there a reason why you think this change is something we should do?

tracing-tracy/src/lib.rs
```diff
@@ -121,14 +124,13 @@ impl<F> TracyLayer<F> {
...
-        let mut max_len =
-            usize::from(u16::max_value()) - 2 - 4 - 4 - function.len() - 1 - file.len() - 1;
+        let mut max_len = usize::from(u16::MAX) - 2 - 4 - 4 - function.len() - 1 - file.len() - 1;
         if data.len() >= max_len {
             while !data.is_char_boundary(max_len) {
                 max_len -= 1;
             }
             self.client
-                .color_message(error_msg, 0xFF000000, self.stack_depth);
+                .color_message(error_msg, 0xFF00_0000, self.stack_depth);
```
nagisa, line 133: Eh, this is a toss-up. Since this is a colour hex, if we’re making it more readable it would make more sense to either `0xFF0000_00` (the alpha "channel" is ignored) or `0xFF_00_00_00` for the 4 channels in RGBx.

tracing-tracy/src/lib.rs
```diff
@@ -148,64 +150,93 @@ where
...
-        if let Some(span) = ctx.span(id) {
-            let mut extensions = span.extensions_mut();
-            if extensions.get_mut::<FormattedFields<F>>().is_none() {
-                let mut fields = FormattedFields::<F>::new(String::with_capacity(64));
-                if self.fmt.format_fields(fields.as_writer(), attrs).is_ok() {
-                    extensions.insert(fields);
-                }
+        let Some(span) = ctx.span(id) else {
+            return;
+        };
```
nagisa, line 155: I… feel that in cases like these, where `if-let` does not already have a diverging `else` branch, use of `let-else` makes code readability a lil’ bit worse. Even if the semicolon from `return;` were removed so that `rustfmt` collapsed this statement into a single line… it’d be an improvement over the current suggestion, but it is still much easier to notice `if let` than it is to notice the `else { ... }`.

tracing-tracy/src/tests.rs
```diff
@@ -99,7 +99,7 @@ fn span_with_fields() {
     let _enter = span.enter();
 }
 
-pub(crate) fn test() {
+pub fn test() {
```
nagisa, line 102: What’s the motivation for making these `pub`? I generally tend to prefer the rule of thumb of minimal visibility as that gives the compiler and its lints more information to work with (e.g. it can tell you about unused definitions.)

tracy-client-sys/tests.rs
```diff
@@ -1,6 +1,10 @@
 #[cfg(all(feature = "enable", test))]
 mod tests {
-    use tracy_client_sys::*;
+    use tracy_client_sys::{
+        ___tracy_emit_message, ___tracy_emit_zone_begin, ___tracy_emit_zone_end,
+        ___tracy_fiber_enter, ___tracy_shutdown_profiler, ___tracy_source_location_data,
+        ___tracy_startup_profiler,
+    };
```
nagisa, line 7: Eh, with all the downsides of glob imports, this current use of it here seems like it can’t go wrong. Do you have any specific motivation for making this change?

tracy-client/benches/client.rs
```diff
@@ -14,7 +14,7 @@ fn client_clone(c: &mut Criterion) {
 
 fn client_running(c: &mut Criterion) {
     let _client = Client::start();
-    c.bench_function("running", |b| b.iter(|| Client::running()));
+    c.bench_function("running", |b| b.iter(Client::running));
```
nagisa, line 17: Nice! :+1:

tracing-tracy/src/lib.rs
```diff
@@ -84,6 +84,7 @@ impl TracyLayer<DefaultFields> {
     /// Create a new `TracyLayer`.
     ///
     /// Defaults to collecting stack traces.
+    #[must_use]
```
nagisa, line 87: Nice! :+1:

tracing-tracy/src/lib.rs
```diff
@@ -99,12 +100,14 @@ impl<F> TracyLayer<F> {
...
-    pub fn with_stackdepth(mut self, stack_depth: u16) -> Self {
+    #[must_use]
+    pub const fn with_stackdepth(mut self, stack_depth: u16) -> Self {
         self.stack_depth = stack_depth;
         self
     }
 
     /// Use a custom field formatting implementation.
+    #[must_use]
     pub fn with_formatter<Fmt>(self, fmt: Fmt) -> TracyLayer<Fmt> {
```
nagisa, line 111: This can be a `const fn` too, can’t it?

tracy-client/src/gpu.rs
```diff
@@ -156,10 +156,10 @@ impl std::error::Error for GpuSpanCreationError {}
 impl Client {
     /// Creates a new GPU context.
     ///
-    /// - 'name' is the name of the context.
-    /// - 'ty' is the type (backend) of the context.
-    /// - 'gpu_timestamp' is the gpu side timestamp the corresponds (as close as possible) to this call.
-    /// - 'period' is the period of the gpu clock in nanoseconds (setting 1.0 means the clock is 1GHz, 1000.0 means 1MHz, etc).
+    /// - `name` is the name of the context.
```
nagisa, line 159: Nice! :+1:

tracy-client/src/gpu.rs
```diff
@@ -259,10 +259,10 @@ impl GpuContext {
             // always be smaller than u64, so no data will be lost.
             unsafe {
                 sys::___tracy_emit_gpu_zone_begin_serial(sys::___tracy_gpu_zone_begin_data {
-                    srcloc: (&span_location.data) as *const _ as usize as u64,
+                    srcloc: std::ptr::addr_of!(span_location.data) as usize as u64,
```
nagisa, line 262: Nice! :+1:

tracy-client/src/lib.rs
```diff
@@ -51,18 +51,18 @@ pub mod internal {
     use std::ffi::CString;
     pub use std::ptr::null;
 
-    #[inline(always)]
+    #[must_use]
```
nagisa, line 54: Unless there is a good reason not to (e.g. measured performance is worse,) lets keep the  `#[inline(always)]`.

tracy-client/src/lib.rs
```diff
@@ -76,22 +76,21 @@ pub mod internal {
...
         crate::plot::PlotName(name)
     }
 
-    #[inline(always)]
     /// Safety: `name` must be null-terminated, and a `Client` must be enabled
     pub unsafe fn set_thread_name(name: *const u8) {
         #[cfg(feature = "enable")]
         unsafe {
-            sys::___tracy_set_thread_name(name.cast())
+            sys::___tracy_set_thread_name(name.cast());
```
nagisa, line 93: In this case the lack of the semicolon actually serves a good purpose – if an upgrade of the `tracy-client-sys` introduced a return type for this function, it would fail to compile, which is something quite desirable, if only to double check that we can ignore it.

tracy-client/src/span.rs
```diff
@@ -184,7 +186,7 @@ impl Drop for Span {
     }
 }
 
-/// Construct a <code>&'static [SpanLocation]</code>.
+/// Construct a `&'static [SpanLocation]`.
```
nagisa, line 189: I suspect this may be a change in the generated docs from a link to `SpanLocation` to a `[SpanLocation]` without a link.

tracy-client/src/state.rs
```diff
@@ -80,12 +80,12 @@ impl Client {
             let mut old_state = CLIENT_STATE.load(Ordering::Relaxed);
             loop {
                 match old_state {
-                    STATE_ENABLED => return Client(()),
+                    STATE_ENABLED => return Self(()),
```
nagisa, line 83: Nice! :+1:

nagisa (commented):
There are some changes that are quite nice, but also some that I would definitely want some elaboration on *why* the change should be made. I annotated places like these with inline comments. I want to stress that my requests for motivation are by no means a rejection or a push-back against the change in question; rather I just think it would be a good idea to have some text written down to make reasoning explicit (even if it is “just because”) and to reduce the space for assumptions.

As for `let-else`, I suspect it’ll be difficult to convince me of its benefit in the uses presented, but I’m happy to debate and to be shown errors of my ways ;)

tracing-tracy/Cargo.toml
SUPERCILEX: No, this is more of a "we should be up-to-date" axium. The changes are listed in this chapter: https://doc.rust-lang.org/edition-guide/rust-2021/index.html.

_Why not?_ is probably a better question IMO.

tracing-tracy/src/lib.rs
SUPERCILEX: Looks like this comes from my personal clippy config and not the default, so reverting.

tracing-tracy/src/lib.rs
SUPERCILEX: Hmmm, yeah this is personal preference. I'm used to early returns and prefer avoiding nesting, but I'm happy to revert if you'd like to keep the old way.

tracing-tracy/src/tests.rs
SUPERCILEX: My personal clippy config again since they're only cfg(test). Reverting.

tracy-client-sys/tests.rs
SUPERCILEX: My clippy config again. Reverting.

tracing-tracy/src/lib.rs
SUPERCILEX: Interestingly enough, no because self gets dropped:
```
destructor of `TracyLayer<F>` cannot be evaluated at compile-time
```

tracy-client/src/lib.rs
SUPERCILEX: Did you benchmark their use? The default should be to **not** use `inline(always)`: https://rust-lang.github.io/rust-clippy/master/index.html#/inline_always. I've also had folks from the stdlib team tell me that it slows down compile times and bloats codegen if llvm thinks using a function would actually be better. Basically the rule is that you need to have measured a perf improvement before using `inline(always)`.

tracy-client/src/lib.rs
SUPERCILEX: Smart! In that case, let's make it explicit.

tracy-client/src/span.rs
SUPERCILEX: Ah, yeah you're right. I spaced and read this as an array for whatever reason.

tracy-client/src/lib.rs
nagisa: I do recall looking at the assembly way back for some of these. It might very well be the case that after the many LLVM upgrades the situation has changed, or it might be that it hasn’t, I don’t really have the time to look into this right now.

---

That said, I can motivate why these *should* be `#[inline(always)]` even outside of the “measure-and-prove” line of reasoning.

Particularly worth noting is that many of these `internal` functions are just basic wrappers over otherwise private constructors to be used in this crate’s macros. Constructing an ADT is pretty much *always* simpler than making a function call to do the same, no matter how many uses of this operation there are.

This is especially notable in the context of the compile-time blowup reasoning that clippy presents as a counter-argument. It makes a lot of sense when we’re talking about functions with a non-trivial amount of logic and such. It does not hold at all for wrapper functions like these (and clippy itself presents some exceptions such as functions that have nothing in their body or simply panic.)

If it wasn’t for the limitations in the Rust’s visibility system in combination with macros, these functions wouldn’t exist at all. In that sense `#[inline(always)]` serves as a pretty good mechanism to also express that sort of intent.

Finally `inline(always)` has a benefit of achieving these intents even in debug builds.

tracing-tracy/src/lib.rs
nagisa: I don’t particularly mind this change much as long as `else { return; };` becomes `else { return };` and is formatted accordingly.

SUPERCILEX: Sounds good, done!

tracy-client/src/lib.rs
SUPERCILEX: I think that makes sense? It still leaves me uneasy lol. Anyway, I brought back the `inline(always)` attributes for the `mod internal` methods.

Did you want all the inlines brought back? I'll disagree a bit more strongly about the others.

SUPERCILEX:
Alrighty! Unless there's more discussion around `inline(always)`, I think this is ready to go?

BTW sorry for continually adding more commits, I've kinda been using this a dumping ground :grin:.

nagisa:
Yeah, seems good to go. Thank you!

SUPERCILEX:
Woohoo, thanks!

PR #86, opened 2023-12-26
Minimize tracing-tracy allocations

SUPERCILEX:
~~Depends on https://github.com/nagisa/rust_tracy_client/pull/85. Will rebase once that's in.~~ Done.

---

~~Note that the two commits should NOT be squashed as the second one contains benchmarking code.~~

Turns out GitHub [saves the PR history](https://stackoverflow.com/a/56632574/4548500) even when squashing, so nevermind since I feel like these commits are messier.

---

The one major downside of this caching is that for long-running applications which occasionally have super long field values, we have no way of reclaiming that memory. I think that's ok until someone actually runs into this issue and can tell us more about their needs. For example, maybe we offer a "clear cache" API that they could run? Or maybe they need to configure the max cache size? Not sure, so probs easiest to wait unless you think this needs to be solved (in which case I'd rather do that in a follow up PR). TL;DR: the memory usage currently scales as O(num_active_spans*max_field_value_length).

---

The results are in: almost double the performance!

> Note that I'm testing this with https://github.com/nagisa/rust_tracy_client/pull/87 as I'll deal with the `format!` allocation once we have a path forward there. 

```
~> tracy-capture -o ~/Desktop/ftzz4.tracy -f
Connecting to 127.0.0.1:8086...
Queue delay: 19 ns
Timer resolution: 19 ns
 471.09 Mbps / 18.8% =2501.45 Mbps | Tx: 563 MB | 1020.05 MB | 10.21 s
Frames: 2
Time span: 10.25 s
Zones: 4,033,823
Elapsed time: 9.93 s
Saving trace... done!
Trace size 282.4 MB (37.28% ratio)
~> tracy-capture -o ~/Desktop/ftzz5.tracy -f
Connecting to 127.0.0.1:8086...
Queue delay: 21 ns
Timer resolution: 21 ns
 498.87 Mbps / 18.2% =2744.67 Mbps | Tx: 366.32 MB | 722.33 MB | 6.67 s
Frames: 2
Time span: 6.69 s
Zones: 4,033,813
Elapsed time: 6.41 s
Saving trace... done!
Trace size 193.94 MB (34.32% ratio)
```

Before:
image

After:
image

### DIFF
--- a/tracing-tracy/src/lib.rs
+++ b/tracing-tracy/src/lib.rs
@@ -49,6 +49,7 @@
 #![doc = include_str!("../FEATURES.mkd")]
 #![cfg_attr(tracing_tracy_docs, feature(doc_auto_cfg))]
 
+use std::cell::Cell;
 use std::{borrow::Cow, cell::RefCell, collections::VecDeque, fmt::Write};
 use tracing_core::{
     field::{Field, Visit},
@@ -176,11 +177,16 @@ where
     }
 
     fn on_event(&self, event: &Event, _: Context<'_, S>) {
+        thread_local! {
+            static BUF: Cell<String> = Cell::new(String::new());
+        }
+
         let mut visitor = TracyEventFieldVisitor {
-            dest: String::with_capacity(64),
+            dest: BUF.take(),
             first: true,
             frame_mark: false,
         };
+
         event.record(&mut visitor);
         if !visitor.first {
             self.client.message(
@@ -196,6 +202,9 @@ where
         if visitor.frame_mark {
             self.client.frame_mark();
         }
+
+        visitor.dest.clear();
+        BUF.set(visitor.dest)
     }
 
     fn on_enter(&self, id: &Id, ctx: Context<S>) {
@@ -273,11 +282,11 @@ impl Visit for TracyEventFieldVisitor {
         // FIXME: this is a very crude formatter, but we don’t have
         // an easy way to do anything better...
         if self.first {
-            let _ = write!(&mut self.dest, "{} = {:?}", field.name(), value);
             self.first = false;
         } else {
-            let _ = write!(&mut self.dest, ", {} = {:?}", field.name(), value);
+            self.dest.push_str(", ");
         }
+        let _ = write!(&mut self.dest, "{} = {value:?}", field.name());
     }
 }
 
--- a/tracing-tracy/src/lib.rs
+++ b/tracing-tracy/src/lib.rs
@@ -49,6 +49,7 @@
 #![doc = include_str!("../FEATURES.mkd")]
 #![cfg_attr(tracing_tracy_docs, feature(doc_auto_cfg))]
 
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::{borrow::Cow, fmt::Write, mem};
 use tracing_core::{
     field::{Field, Visit},
@@ -142,6 +143,22 @@ impl Default for TracyLayer {
     fn default() -> Self {
         Self::new()
     }
+}
+
+static MAX_CACHE_SIZE: AtomicUsize = AtomicUsize::new(usize::MAX);
+
+/// Specify the maximum number of bytes used in thread local caches.
+///
+/// A value of zero disables the cache, while a value of [`usize::MAX`] denotes an unlimited
+/// cache size.
+///
+/// Note: the upper bound on the cache size is respected on a best effort basis only. We make
+/// no guarantees on the maximum memory used by tracing-tracy. Notably, changes to this value
+/// are eventually consistent, i.e. caches are not flushed upon an update.
+///
+/// Defaults to [`usize::MAX`].
+pub fn set_max_cache_size(max_bytes_used_per_thread: usize) {
+    MAX_CACHE_SIZE.store(max_bytes_used_per_thread, Ordering::Relaxed);
 }
 
 thread_local! {
@@ -318,9 +335,12 @@ fn main() {
 }
 
 mod utils {
-    use std::cell::UnsafeCell;
+    use crate::MAX_CACHE_SIZE;
+    use std::cell::{Cell, UnsafeCell};
     use std::mem;
+    use std::mem::ManuallyDrop;
     use std::ops::{Deref, DerefMut};
+    use std::sync::atomic::Ordering;
 
     pub struct VecCell<T>(UnsafeCell<Vec<T>>);
 
@@ -338,23 +358,47 @@ mod utils {
         }
     }
 
-    pub struct StrCache(VecCell<String>);
+    pub struct StrCache {
+        str_bufs: VecCell<String>,
+        total_size: Cell<usize>,
+    }
 
     impl StrCache {
         pub const fn new() -> Self {
-            Self(VecCell::new())
+            Self {
+                str_bufs: VecCell::new(),
+                total_size: Cell::new(0),
+            }
         }
 
         pub fn acquire(&self) -> StrCacheGuard {
             StrCacheGuard::new(
                 self,
-                self.0.pop().unwrap_or_else(|| String::with_capacity(64)),
+                self.str_bufs
+                    .pop()
+                    // TODO use inspect once 1.76 is stable
+                    .map(|buf| {
+                        self.total_size.set(self.total_size.get() - buf.capacity());
+                        buf
+                    })
+                    .unwrap_or_else(|| String::with_capacity(64)),
             )
         }
 
         fn release(&self, mut buf: String) {
+            let new_cache_size = self.total_size.get().saturating_add(buf.capacity());
+            if new_cache_size == usize::MAX {
+                // This is never going to happen, but if we've used the entire address space,
+                // don't bother adding another cache entry as this keeps the logic simpler.
+                return;
+            };
+            if buf.capacity() == 0 || new_cache_size > MAX_CACHE_SIZE.load(Ordering::Relaxed) {
+                return;
+            }
+            self.total_size.set(new_cache_size);
+
             buf.clear();
-            self.0.push(buf);
+            self.str_bufs.push(buf);
         }
     }
 
@@ -368,8 +412,9 @@ mod utils {
             Self { cache, buf }
         }
 
-        pub fn into_inner(mut self) -> String {
-            mem::take(&mut self.buf)
+        pub fn into_inner(self) -> String {
+            let mut this = ManuallyDrop::new(self);
+            mem::take(&mut this.buf)
         }
     }
 
--- a/tracing-tracy/src/lib.rs
+++ b/tracing-tracy/src/lib.rs
@@ -145,7 +145,7 @@ impl Default for TracyLayer {
     }
 }
 
-static MAX_CACHE_SIZE: AtomicUsize = AtomicUsize::new(usize::MAX);
+static MAX_CACHE_SIZE: AtomicUsize = AtomicUsize::new(8192);
 
 /// Specify the maximum number of bytes used in thread local caches.
 ///
@@ -156,7 +156,7 @@ static MAX_CACHE_SIZE: AtomicUsize = AtomicUsize::new(usize::MAX);
 /// no guarantees on the maximum memory used by tracing-tracy. Notably, changes to this value
 /// are eventually consistent, i.e. caches are not flushed upon an update.
 ///
-/// Defaults to [`usize::MAX`].
+/// Defaults to `8192`.
 pub fn set_max_cache_size(max_bytes_used_per_thread: usize) {
     MAX_CACHE_SIZE.store(max_bytes_used_per_thread, Ordering::Relaxed);
 }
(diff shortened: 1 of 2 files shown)


SUPERCILEX:
Ok, this should be ready for review. I've updated the PR description with results.

tracing-tracy/src/lib.rs
```diff
@@ -154,7 +177,8 @@ where
 
         let mut extensions = span.extensions_mut();
         if extensions.get_mut::<FormattedFields<F>>().is_none() {
-            let mut fields = FormattedFields::<F>::new(String::with_capacity(64));
+            let mut fields =
+                FormattedFields::<F>::new(CACHE.with(|cache| cache.pop().unwrap_or_default()));
```
nagisa, line 181: Consider adding a newtype that pushes contained string back into the cache on drop. As written if `format_fields` is not `Ok`, the buffer will be returned to heap.

tracing-tracy/src/lib.rs
```diff
@@ -144,6 +163,10 @@ impl Default for TracyLayer {
     }
 }
 
+thread_local! {
+    static CACHE: VecCell<String> = const { VecCell::new() };
```
nagisa, line 167: FWIW I had great success using `crossbeam_queue::ArrayQueue` for this sort of thing in the past. It may be somewhat slower on the account that it is a global queue (and thus there is going to cache lane contention and whatnot.) Benefit is that you no longer need to worry about situations where a span is created in one thread, moved into another and dropped, or worry about `!Send`/`!Sync` issues, as the `ArrayQueue` is both.

And if `ArrayQueue` turns out to be a poor fit for this use-case, then this `CACHE` definitely needs at least an upper bound on how many elements it can store here.

tracing-tracy/src/lib.rs
```diff
@@ -228,40 +262,56 @@ where
...
+                "Exiting a tracing span, but got nothing on the tracy span stack!",
+                0xFF000000,
+                self.stack_depth,
+            );
+        }
+    }
+
+    fn on_close(&self, id: Id, ctx: Context<'_, S>) {
+        let Some(span_data) = ctx.span(&id) else {
+            return;
```
nagisa, line 297:
```suggestion
            return
```

tracing-tracy/src/lib.rs
```diff
@@ -228,40 +262,56 @@ where
...
+        };
+
+        if let Some(fields) = span_data.extensions_mut().get_mut::<FormattedFields<F>>() {
```
nagisa, line 300: At this point use `let-else` too?

tracing-tracy/src/lib.rs
nagisa: I also feel like it would perhaps make sense to keep `with_capacity` to reduce reallocations for the very first strings used here. Admittedly `64` was a guesstimate, but it does save on potentially, like, 5 reallocs?

tracing-tracy/src/lib.rs
```diff
@@ -273,11 +323,11 @@ impl Visit for TracyEventFieldVisitor {
         // FIXME: this is a very crude formatter, but we don’t have
         // an easy way to do anything better...
         if self.first {
-            let _ = write!(&mut self.dest, "{} = {:?}", field.name(), value);
             self.first = false;
         } else {
-            let _ = write!(&mut self.dest, ", {} = {:?}", field.name(), value);
+            self.dest.push_str(", ");
         }
+        let _ = write!(&mut self.dest, "{} = {value:?}", field.name());
```
nagisa, line 330: I believe the original approach gives the formatting infrastructure a way to do the same thing with potentially one less (re-)allocation, doesn’t it?

tracing-tracy/src/lib.rs
SUPERCILEX: Whoops, great catch thanks! I added a guard type and brought back the 64 byte allocation as that does save a few hundred reallocs in my testing.

tracing-tracy/src/lib.rs
SUPERCILEX: That's a pretty clever idea! However, I'm fairly opposed to adding a central contention point for all threads. Given that Tracy claims to be a nanosecond resolution tracer, I'm pretty sure we'd be losing that claim with a shared cache line. Or at the very least, I would want to see benchmarks of trying to write to an atomic from 16 threads simultaneously to see what the latency looks like.

tracing-tracy/src/lib.rs
SUPERCILEX: Sadly this doesn't format property. I think they limit the line length to 80 chars to get the one-liner.

tracing-tracy/src/lib.rs
SUPERCILEX: Yeah, I can't tell without stepping through the assembly which is better, so I went back to the old version: https://rust.godbolt.org/z/W7Y8vdxzb

I hand-wrote the str and bool implementations since they can be done without the fmt infrastructure and with a single allocation.

tracing-tracy/src/lib.rs
```diff
@@ -144,6 +145,26 @@ impl Default for TracyLayer {
...
+/// Specify the maximum number of bytes used in thread local caches.
+///
+/// A value of zero disables the cache, while a value of [`usize::MAX`] denotes an unlimited
+/// cache size.
+///
+/// Note: the upper bound on the cache size is respected on a best effort basis only. We make
+/// no guarantees on the maximum memory used by tracing-tracy. Notably, changes to this value
+/// are eventually consistent, i.e. caches are not flushed upon an update.
+///
+/// Defaults to [`usize::MAX`].
```
nagisa, line 159: Hm, `usize::MAX` is a poor default, I feel. As I said before, it is pretty straightforward to use `tracing` in a way that would lead people to indefinitely keep allocating strings in one of the threads and then release them in another. I would also argue that people by default won’t go invoking this function to set a finite limit proactively when integrating `tracing-tracy` into their codebase. When that happens, it can be pretty difficult to actually troubleshoot and figure out where the memory is getting leaked to when it starts happening (most likely in production.)

On the other hand "tracing is not going fast enough" is much easier to investigate (you can easily narrow it down with some directed experiments.)

tracing-tracy/src/lib.rs
```diff
@@ -228,55 +257,86 @@ where
...
     first: bool,
 }
 
-impl Visit for TracyEventFieldVisitor {
+impl Visit for TracyEventFieldVisitor<'_> {
     fn record_bool(&mut self, field: &Field, value: bool) {
         match (value, field.name()) {
-            (true, "tracy.frame_mark") => self.frame_mark = true,
-            _ => self.record_debug(field, &value),
+            (_, "tracy.frame_mark") => self.frame_mark = value,
```
SUPERCILEX, line 309: I changed this to set the value... was the previous behavior intentional?

tracing-tracy/src/lib.rs
nagisa: I also am left wondering if there’s much value in trying to track the size/capacity of the strings contained within the cache (rather than just a number of elements.)

One reason why I’m saying this is because if the code spawns N threads, `tracing-tracy` will ultimately end up using N*bytes space, and N is largely unbounded.

And then tracking bytes has a potential issue in that if the cache ends up with a single very large string inside it, it can end up pushing out many small strings out, which would end up likely hurting the throughput much more.

SUPERCILEX: > Hm, usize::MAX is a poor default

Fair enough, how about 8K?

> allocating strings in one of the threads and then release them in another

Hmmm, hadn't thought of this. Wouldn't it be pretty unusual to never re-use allocations on the receiver thread though? Here's how I'm thinking spans can move across threads:
- Async: each await point leads to potential thread switching.
- Spawn a background task and move the span into the closure. This would be the case where you don't get to re-use allocations, but it seems weird to me that you'd explicitly start a span on one thread and move into another since your timing between the task being scheduled and starting will be pretty arbitrary.
- Anything else I'm missing?

If we want to focus on the background task case (which I'm not sure we should unless it's more common than I think), then that's where using your ArrayQueue seems like a great idea.

> I also am left wondering if there’s much value in trying to track the size/capacity of the strings contained within the cache (rather than just a number of elements.)

Yeah, I spent some time thinking about this too and didn't really get anywhere. I'm not sure I agree that N is unbounded in practice. I can see how thread counts could increase burstly, but when the thread dies its TLS gets de-allocated so we're fine.

My primary concern with simply using the number of strings is that if you have a single span that has HUGE fields for whatever reason, you'll be stuck with this massive allocation forever whereas the max bytes approach will probably kill it.

> And then tracking bytes has a potential issue in that if the cache ends up with a single very large string inside it, it can end up pushing out many small strings out

Yup, this is really annoying and bothered me too. My best idea to fix this is to sort the vec by string size every time we trim it. But then I wasn't sure which strings to get rid of... nuking small strings means one very large string can dominate while nuking large strings means you lose end up realloc-ing more. If you think this is a good idea, I think the best approach is probably nuking the biggest strings since that will bias towards having a lot of medium sized strings (I think).

nagisa: Yeah, in async scenarios you are almost definitely not going to have the same allocation but it is also quite guaranteed that there is going to be a reasonably good random distribution of which thread either end gets executed by.

The 2nd case with "spawning a background task" is the concerning one. I actually do something quite like it at $DAYJOB (though moving into threads of rayon pool) in order to establish parent-child relationship between spans of the parent thread and the work that gets executed on the rayon pool. Otherwise the span gets "detached" and the stuff that happens within rayon is not as actionable.

Perhaps something that’s unique to rayon threads is that work that happens within those is largely "scoped". I.e. work that happens within is guaranteed to start and finish within the boundaries of the parent span, only that there can be multiple parallel ones. But it is also a perfect example of where the movement of span ownership is very directional.

Here's a [basic example](https://github.com/near/nearcore/blob/4942ba29f295f7e7b6e452c84000bb48aee79262/tools/mock-node/src/setup.rs#L199)

SUPERCILEX: I'm a little confused, in the example provided isn't the parent span allocated and de-allocated on the same thread since it waits for all the tasks to be finished? Seems like it'd work perfectly in this scenario.

nagisa: Ah right, my bad. This is not `move` closure. I did distinctly remember them being passed around by value and such, but the code must have changed since. And indeed scoped threads can use references if they wish to do so.

So the concerning part is really just moving a span into a different (possibly non-scoped) thread. How likely is that kinda use? Not sure. Is it possible? Yes, definitely. Are there use-cases that doing so serves? Context.

SUPERCILEX: > moving a span into a different (possibly non-scoped) thread. How likely is that kinda use?

I feel like pretty low. Or at least this code doesn't make that case much worse (it'll take up extra space allocating the vec but that's it). If we wanted to avoid any downside from this code, we'd need to add a max thread local bytes used API and then setting it to 0 makes these changes behavior the same as before. We totally could, but I feel like it would be a pretty niche use case. Do you want me to add that API, or should we wait for someone who needs it?

nagisa: My only requirement is that there isn’t an infinite memory leak (or unsoundness, or...) for that kind of code. I’m okay with things not being perfect, say, perf-wise for likely rare corner-cases like these.

SUPERCILEX: Ah, cool cool! Then let me go with the 8K default limit and biggest string eviction policy.

SUPERCILEX: Done.

tracing-tracy/src/lib.rs
nagisa: In principle a field name for an event can occur only once. I don’t particularly care either way, but `tracy.frame_mark = false` does not really have any reason to show up in the user code… so it seemed to make sense to not invoke any special logic on it? Unless I guess they want to make it dynamic somehow? :shrug:

tracing-tracy/src/lib.rs
```diff
@@ -228,55 +257,86 @@ where
...
+            (true, _) => self.record_str(field, "true"),
+            (false, _) => self.record_str(field, "false"),
         }
     }
 
+    fn record_str(&mut self, field: &Field, value: &str) {
+        {
```
nagisa, line 316: Why the extra block? I would also extract `field.name()` into a variable.

tracing-tracy/src/lib.rs
```diff
@@ -291,3 +351,117 @@ fn main() {
...
+
+    pub struct VecCell<T>(UnsafeCell<Vec<T>>);
+
+    impl<T> VecCell<T> {
+        pub const fn new() -> Self {
+            Self(UnsafeCell::new(Vec::new()))
+        }
+
+        pub fn push(&self, item: T) {
+            unsafe { &mut *self.0.get() }.push(item);
```
nagisa, line 371: Please write down safety comments (i.e. why this unsafe operation is guaranteed to be correct / how are the preconditions imposed by the callee are being satisfied.)

nagisa (approved):
In principle LGTM, last few nits and comments inline.

tracing-tracy/src/lib.rs
SUPERCILEX: > Unless I guess they want to make it dynamic somehow?

Yeah that's all I could think of. It just felt wrong to me to take in a bool and ignore it. I could see it as being confusing behavior to track down if you're expecting false to work for whatever reason.

tracing-tracy/src/lib.rs
SUPERCILEX: Trying to scope the `alloc_always_size` variable, but eh kinda unnecessary. Done.

tracing-tracy/src/lib.rs
SUPERCILEX: Done.

Issue #78, opened 2023-12-20
Client never calls `___tracy_shutdown_profiler`


jimblandy:
`tracy_client::Client` doesn't provide any way to call `___tracy_shutdown_profiler`, which leads to incomplete traces and complaints from the Tracy GUI:

![image](https://github.com/nagisa/rust_tracy_client/assets/751272/f95b0a78-22b1-4624-9fda-0740599f82a3)

I gather from other issues that there was previously some reference counting magic somewhere but it got ripped out, so I can believe it's difficult to tell when it is safe to call the shutdown function. But it does seem like not cleaning things up causes problems.

I'd love to hear more about the background than I could glean from reading `tracy-client/src/state.rs`.

nagisa:
https://github.com/nagisa/rust_tracy_client/blob/53fa375971520bfd2e9c8fd80f292ed64d15c263/tracy-client/src/state.rs#L27-L28

is the relevant comment. I don't recall seeing this message before, but since you mention problems, I guess a question worth asking is whether there are truly missing source locations, unresolved stack frames, etc, that you're seeing?

Regardless it seems quite plausible to pursue implementing a dedicated lifetime management thread if only to get rid of this message. I can’t imagine an use-case for setting up and tearing down the client frequently enough that the thread setup costs would become unbearable.

jimblandy:
I don't know, exactly. I was profiling `wgpu`'s "water" example, which was reaching 700FPS on my system and spewing out quite a lot of data. In the end more than half of the profile was received by Tracy after I hit ESC to exit the demo.

Looking at the Tracy v0.10 sources (`tracy-client-sys` v0.22), it seems like none of this should matter if you don't select the `delayed-init` and `manual-lifetime` features. Without those, it seems like Tracy v0.10 will take care of all this initialization itself. Would it make sense for `tracy-client` to simply not request those features?

nagisa:
https://github.com/nagisa/rust_tracy_client/issues/29 is the relevant reading for why `manual-lifetime` is required.

jimblandy:
I see. One last question: `state.rs` says:

> Alas, Tracy's extensive use of thread-local storage presents us with another problem – we must start up and shut down the client within the same thread.

How does this requirement arise? I see that `tracy::ShutdownProfiler` does use some `thread_local` variables, but not in a way that would cause a problem if `StartupProfiler` and `ShutdownProfiler` were called on different threads.

nagisa:
[...]

PR #88, opened 2024-01-01
Turn configuration options into const generic params

SUPERCILEX:
Depends on https://github.com/nagisa/rust_tracy_client/pull/87.

This is a bit of a _so preoccupied with whether they could, they didn’t stop to think if they should_ moment, but I realized we can make our primitive config options const generic params. The benefit is that the optimizer can eliminate a bunch of branches and dead code that would have been very hard to prove safe to remove before. There's precedent for doing something similar in other crates: https://docs.rs/sharded-slab/0.1.7/sharded_slab/pool/struct.Pool.html#method.new_with_config.

Also, I don't think this makes the API worse, it's basically just different syntax (unless you want this stuff to be dynamic in which case you're screwed):
```rust
   tracing_tracy::TracyLayer::new()
                        .with_stack_depth::<32>()
                        .with_fields_in_zone_name::<false>(),
```

Thoughts?

### DIFF
--- a/tracing-tracy/src/lib.rs
+++ b/tracing-tracy/src/lib.rs
@@ -66,6 +66,7 @@ use tracing_subscriber::{
 use client::{Client, Span};
 use utils::{StrCache, StrCacheGuard, VecCell};
 
+use crate::config::{ConfigOption, True, ZeroU16};
 pub use client;
 
 thread_local! {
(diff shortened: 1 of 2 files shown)


nagisa:
> Thoughts?

NACK if its an either/or ordeal. I know there are folks who configure their layers dynamically. But. We could make an API allow both dynamic and static configuration:

```rust
TracyLayer::<S, Z>::with_stack_depth<NewS: StackDepthOption>(s: NewS);
TracyLayer::<S, Z>::with_fields_in_zone_name<NewZ: FieldsInZoneNameOption>(s: NewZ);

impl StackDepthOption for u16 { /* dynamic */ }
impl StackDepthImpl for typenum::N0 { /* static */ }
impl StackDepthImpl for typenum::N1 { /* static */ }
// ...

impl FieldsInZoneNameOption for bool { /* dynamic */ }
struct True;
impl FieldsInZoneNameOption for True { /* static */ }
struct False;
impl FieldsInZoneNameOption for False { /* static */ }
```

etc. Or it could be a trait with methods. If those methods return constant values, I suspect the compiler will be able to figure inlining them in.

SUPERCILEX:
Oooooh, that's quite clever! I went forward with that approach, but only differentiated between zero and non-zero u16s since the main optimizable difference for the stack depth param.

nagisa:
Hm, yeah, I’m not sure this is worth it in this exact proposed shape. I’m going to cut a release shortly with all the changes so far and perhaps draft a counter proposal to this...

SUPERCILEX:
That works! I've been procrastinating on replying to #78, so I'll try and do that tonight. The main point is that I agree we should probably temporarily go back to not manual lifetimes until we figure out a proper fix.

nagisa:
cf. #91

SUPERCILEX:
#91 is better than this as long as `ConstConfig` or something equivalent is provided out-of-the box.

PR #91, opened 2024-01-06
tracing-tracy: replace baked-in config fields with a `Config` trait

nagisa:
This allows users to more flexibly adjust the behaviour of the tracing layer. My favourite examples are actually an ability to implement this structure for some other serialization format that holds the application-wide config. That way people do not need to litter their application set up code with accesses to their configuration, conversion and other concerns.

But with this if people want constant evaluation for performance critical applications, now they can do it too.

### DIFF
--- a/tracing-tracy/src/config.rs
+++ b/tracing-tracy/src/config.rs
@@ -1,34 +1,46 @@
 use client::Client;
 use tracing_subscriber::fmt::format::DefaultFields;
 use tracing_subscriber::fmt::FormatFields;
+#[allow(unused_imports)] // for documentation.
+use super::TracyLayer;
 
 /// Configuration of the [`TracyLayer`] behaviour.
 ///
-/// For most users [`DynamicConfig`] is going to be a good default option, however advanced users
-/// can implement this trait manually to achieve better performance through constant evaluation,
-/// to override the formatter used or to otherwise modify the behaviour of `TracyLayer` in ways
-/// that are not exposed via the `DynamicConfig` type.
+/// For most users [`DefaultConfig`] is going to be a good default choice, however advanced users
+/// can implement this trait manually to override the formatter used or to otherwise modify the
+/// behaviour of the `TracyLayer`.
 ///
 /// # Examples
 ///
-/// ## Implementation with compile-time configuration
-///
 /// ```
-/// #[derive(Default)]
-/// struct ConstantTracyConfig {
-///     formatter: tracing_subscriber::fmt::format::DefaultFields,
-/// }
+/// use tracing_subscriber::fmt::format::DefaultFields;
 ///
-/// impl tracing_tracy::Config for ConstantTracyConfig {
-///     type Formatter = tracing_subscriber::fmt::format::DefaultFields;
-///     fn formatter(&self) -> &Self::Formatter { &self.formatter }
-///     fn stack_depth(&self) -> u16 { 0 } // Same as the default trait impl.
-///     fn format_fields_in_zone_name(&self) -> bool { true } // Same as the default trait impl.
+/// struct TracyLayerConfig {
+///     fmt: DefaultFields,
+/// }
+/// impl tracing_tracy::Config for TracyLayerConfig {
+///     type Formatter = DefaultFields;
+///     fn formatter(&self) -> &Self::Formatter {
+///         &self.fmt
+///     }
+///     // The boilerplate ends here
+///
+///     /// Collect 32 frames in stack traces.
+///     fn stack_depth(&self) -> u16 {
+///         32
+///     }
+///
+///     /// Do not format in fields into zone names.
+///     fn format_fields_in_zone_name(&self) -> bool {
+///         false
+///     }
+///
+///     // etc.
 /// }
 /// ```
 ///
-/// With this sort of setup the compiler will be able to inline calls to `stack_depth` and
-/// `format_fields_in_zone_name` and optimize accordingly.
+/// With this configuration `TracyLayer` will collect some call stacks and the formatting of the
+/// zone names is different from the `DefaultConfig`.
 pub trait Config {
     type Formatter: for<'writer> FormatFields<'writer> + 'static;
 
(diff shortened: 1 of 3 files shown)


nagisa:
In the 2nd commit I went ahead and added an example of how `Config` trait enables customization beyond what was reasonable with any sort of field-based approach (`const` or not.)

tracing-tracy/src/config.rs
```diff
@@ -0,0 +1,145 @@
...
+/// }
+/// ```
+///
+/// With this sort of setup the compiler will be able to inline calls to `stack_depth` and
+/// `format_fields_in_zone_name` and optimize accordingly.
+pub trait Config {
+    type Formatter: for<'writer> FormatFields<'writer> + 'static;
+
+    /// Use a custom field formatting implementation.
+    fn formatter(&self) -> &Self::Formatter;
```
SUPERCILEX, line 36: I wish there was a way to specify a default here. Sadly not much progress: https://rust-lang.github.io/rfcs/2532-associated-type-defaults.html

tracing-tracy/src/config.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// Specify the maximum number of stack frames that will be collected.
+    ///
+    /// Note that enabling callstack collection can and will introduce a non-trivial overhead at
+    /// every instrumentation point. Specifying 0 frames (which is the default) will disable stack
```
SUPERCILEX, line 41: Remove "(which is the default)" since that's redundant.

tracing-tracy/src/config.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// Another example a limitation in `Tracy` where the message length or zone name cannot exceed
+    /// a certain (low) limit of bytes.
+    ///
+    /// Although `tracing_tracy` does it best to paper over these sorts of differences, it can’t
+    /// always make them invisible. In certain cases detecting these sorts of issues is
+    /// straightforward, and it is when `tracing_tracy` will invoke this method to enable users to
+    /// report the issues in whatever way they wish to.
+    ///
+    /// By default a message coloured in red is emitted to the tracy client.
+    fn on_error(&self, client: &Client, error: &'static str) {
```
SUPERCILEX, line 77: Ooooooh, this is quite a nice flexibility improvement.

tracing-tracy/src/config.rs
```diff
@@ -0,0 +1,145 @@
...
+            fmt: DefaultFields::new(),
+            stack_depth: 0,
+            fields_in_zone_name: true,
+        }
+    }
+
+    /// Specify the maximum number of stack frames that will be collected.
+    ///
+    /// Note that enabling callstack collection can and will introduce a non-trivial overhead at
+    /// every instrumentation point. Specifying 0 frames (which is the default) will disable stack
```
SUPERCILEX, line 106: Same here.

tracing-tracy/src/lib.rs
```diff
@@ -196,19 +162,28 @@ thread_local! {
...
     fn on_new_span(&self, attrs: &Attributes<'_>, id: &Id, ctx: Context<'_, S>) {
         let Some(span) = ctx.span(id) else { return };
 
         let mut extensions = span.extensions_mut();
-        if extensions.get_mut::<FormattedFields<F>>().is_none() {
-            let mut fields =
-                FormattedFields::<F>::new(CACHE.with(|cache| cache.acquire().into_inner()));
-            if self.fmt.format_fields(fields.as_writer(), attrs).is_ok() {
+        if extensions
+            .get_mut::<FormattedFields<C::Formatter>>()
```
SUPERCILEX, line 175: Are type alises inside impls a thing? Would be nice for `FormattedFields<C::Formatter>`.

tracing-tracy/src/config.rs
```diff
@@ -0,0 +1,145 @@
...
+    fn on_error(&self, client: &Client, error: &'static str) {
+        client.color_message(error, 0xFF000000, self.stack_depth());
+    }
+}
+
+/// A type that implements the [`Config`] trait with runtime-adjustable values.
+///
+/// Ues the [`tracing_subscriber`] [`DefaultFields`] formatter. If not appropriate, consider
+/// implementing the `Config` trait yourself.
+pub struct DynamicConfig {
```
SUPERCILEX, line 86: Can we also provide a `ConstConfig` that looks like my original PR (so purley generic)? I think that would cover the two extremes. Anything in the middle is going to be pretty application specific, so then it makes sense to require implementing a trait. The benefit of providing `ConstConfig` out of the box is that we can use it and make it the default.

tracing-tracy/src/config.rs
nagisa: I wish it was possible too, however even with ATDs it wouldn’t be possible as the method returns a reference to `&self` and traits do not have access to fields. Achieving this would need something like https://github.com/rust-lang/rfcs/pull/1546.

tracing-tracy/src/config.rs
nagisa: I have a feeling that `ConstConfig` is better off defined by the users. While it is possible to define one right now for just `StackDepth` and `FieldsInZoneName` today, this type cannot be expanded in a semver compatible manner as additional configuration options are introduced, leaving it in a spot where:

* it stands out as weirdly incomplete; and
* it isn’t valuable to expand it if there isn’t an upcoming breaking change; and
* any other breaking changes do not force updating of this structure, so it remains incomplete despite the opportunity to bring it up-to-date.

Just to give an example, I have been thinking it might be nice to introduce an ability to add another method to allow computing stack_depth based on `&tracing::Metadata` argument; perhaps even various specializations such as `event_stack_depth`, `span_stack_depth` etc. I’m not yet sure if its a good idea or not, but regardless for dynamic config introducing these in a backward-compatible manner is straightforward; the current methods can pass through configuration to the more fine-grained options to achieve the old effect. For constant config adding, removing or changing a generic in any way that was not coded in the original implementation is a semver break.

To be honest I even pondered if adding a `DynamicConfig` is worth it (over just a `DefaultConfig` that does not provide any way to change its behaviour.) I personally found that whenever a crate provides both options, I eventually end up having my hand forced to migrate to the trait-based configuration anyway.

SUPERCILEX: > I have been thinking it might be nice to introduce an ability to add another method to allow computing stack_depth based on &tracing::Metadata argument; perhaps even various specializations such as event_stack_depth, span_stack_depth etc.

Ah, yeah that's a compelling reason to exclude a ConstConfig.

> To be honest I even pondered if adding a DynamicConfig is worth it (over just a DefaultConfig that does not provide any way to change its behaviour.) I personally found that whenever a crate provides both options, I eventually end up having my hand forced to migrate to the trait-based configuration anyway.

I'd be onboard with this. It'll force everybody (including us) to bias towards const values rather than slower defaults. Of course the downside is that the API a little more annoying to use.

nagisa: I pushed up a commit with a removal of `DynamicConfig` (replaced by a `DefaultConfig` that just uses default trait methods for everything.) And made sure to write up an example snippet for people to copy if they need to modify the behaviour in any way.

tracing-tracy/src/config.rs
```diff
@@ -1,34 +1,46 @@
...
+///         &self.fmt
+///     }
+///     // The boilerplate ends here
+///
+///     /// Collect 32 frames in stack traces.
+///     fn stack_depth(&self) -> u16 {
+///         32
+///     }
+///
+///     /// Do not format in fields into zone names.
```
SUPERCILEX, line 33: Typo: remove "in" I think? Sentence is a little weird otherwise.

tracing-tracy/src/config.rs
SUPERCILEX: Looks good! And if people complain, then I guess we can add it back.

moulins: > Just to give an example, I have been thinking it might be nice to introduce an ability to add another method to allow computing `stack_depth` based on `&tracing::Metadata` argument

Any reason to not add the argument on the existing `stack_depth()` method? Having both `fn stack_depth(&self)` and `fn stack_depth_for(&self, _: &tracing::Metadata)` in the API seems suboptimal and kinda confusing.

I'd suggest also adding this argument to `format_fields_in_zone_name`; only having a "global" switch for this can be pretty limiting.

nagisa: Right now perhaps none, since we're making a breaking change anyway, but default methods on a trait like these allow for providing incrementally fancy ways to configure stuff without necessarily exposing every single user to the maximum amount of complexity.

moulins: In this particular case I'd say the trade-off isn't worth it: there isn't much complexity difference between these two signatures:
```rs
fn stack_depth(&self) -> bool { false }
fn stack_depth(&self, _: &tracing::Metadata<'_>) -> bool { false }
```
And in exchange you now have two very similar methods in the trait, and needs to explain in the documentation which one to implement in which case.

PR #92, opened 2024-01-14
client: support `not(manual-lifetime)` mode after all…

nagisa:


tracy-client/src/state.rs
```diff
@@ -1,4 +1,5 @@
 use crate::Client;
+#[cfg(all(feature = "enable", feature = "manual-lifetime"))]
```
SUPERCILEX, line 2: Might be cleaner to use the module delegate pattern for all these CFGs. So create two `mod imp`s and have `struct Client(imp::Client)` so that you can have a single cfg on the mod declaration.

SUPERCILEX:
Anything I can do to help get this moving along? I'd be happy to send a patch for the delegate cfgs.

tracy-client/src/state.rs
nagisa: I doubt the amount of complexity here right now is sufficiently high for a strategy proposed to pay off.

nagisa:
> Anything I can do to help get this moving along?

I just need to not forget about this by the time weekend comes up ^^

SUPERCILEX:
Haha, sounds good!

tracy-client/src/state.rs
```diff
@@ -161,14 +65,113 @@ impl Clone for Client {
     }
 }
 
-#[cfg(all(test, feature = "enable"))]
-mod test {
-    use super::*;
+#[cfg(all(feature = "enable", feature = "manual-lifetime"))]
+mod manual_lifetime {
```
SUPERCILEX, line 69: Nice! This is the gist of what I meant by a cfged imp module.

PR #121, opened 2024-09-15
Add support for TRACY_DEBUGINFOD

### LINKED ISSUE #120: Support for TRACY_DEBUGINFOD
There doesn't seem to be a way to enable the TRACY_DEBUGINFOD feature.

mahkoh:
Closes #120

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,6 +34,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - run: sudo apt-get update
+      - run: sudo apt-get -y install libdebuginfod-dev
       - run: rustup install nightly --profile=minimal
       - run: rustup default nightly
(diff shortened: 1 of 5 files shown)


nagisa:
Thanks!

nagisa:
Hm, looks like the doc build fails -- something needs to be adjusted to make the header available...

mahkoh:
Added the package to the workflow.

mahkoh:
But neither of those will work on windows. Not sure how to handle this.

mahkoh:
This should work now. The feature is not being tested so there is no need to install the package in that workflow.

nagisa:
Do you know if docs.rs has this package installed? If not, we need to adjust the configuration for docs.rs too.

nagisa:
Looks like not: https://github.com/rust-lang/crates-build-env/blob/master/linux/packages.txt

mahkoh:
Should be handled via https://github.com/rust-lang/docs.rs/issues/147#issuecomment-648844656

nagisa (approved):

PR #125, opened 2024-12-05
Add Client::plot_config + relevant types + example

Dinnerbone:
The example produces this trace with custom plot types:
image


tracy-client/src/plot.rs
```diff
@@ -6,6 +6,61 @@ use crate::Client;
...
+    Stepped,
+
+    /// Lines will be smooth (interpolating a line between two values).
+    #[default]
+    Smooth,
+}
+
+/// Configuration for how a plot appears in the Tracy profiling UI.
+#[derive(Clone, PartialEq, Debug)]
+pub struct PlotConfiguration {
```
Ralith, line 39: Should this be `#[non_exhaustive]`, or maybe even a builder struct with private fields, for forwards compatibility?

Dinnerbone: Builder sounds good! I'll make it so.

Dinnerbone: How's that?

Ralith: LGTM, thanks!

tracy-client/src/plot.rs
```diff
@@ -6,6 +6,89 @@ use crate::Client;
...
+pub enum PlotLineStyle {
+    /// Lines will be stepped (ie look like a staircase).
+    Stepped,
+
+    /// Lines will be smooth (interpolating a line between two values).
+    #[default]
+    Smooth,
+}
+
+/// Configuration for how a plot appears in the Tracy profiling UI.
```
nagisa, line 39: Worth adding a code example here. I found it to be pretty common for people to struggle to figure out how to construct stuff, when the construction is not a `fn new()` (even if it is as simple as using `default()` instead.) A plain example would help with that.

nagisa (approved):

nagisa:
This is all in 0.17.5.

PR #130, opened 2024-12-25
Allow GPU timestamps to be emitted in monotonically-increasing order, as Tracy requires.

pcwalton:
Tracy gets confused if you emit GPU timestamps in anything other than monotonically-increasing order. For example, nested spans *outer* and *inner*, you must supply the timestamps in this order: (1) *outer* start; (2) *inner* start; (3) *inner* end; (4) *outer* end. Following this restriction for nested spans isn't possible with the current API, because it requires the client to supply the start and end timestamps at the same time. So I changed the API to supply the start and end timestamps separately and documented Tracy's restriction.

### DIFF
--- a/tracy-client/src/gpu.rs
+++ b/tracy-client/src/gpu.rs
@@ -76,8 +76,6 @@ pub struct GpuContext {
     #[cfg(feature = "enable")]
     value: u8,
     #[cfg(feature = "enable")]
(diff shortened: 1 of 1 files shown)


cwfitzgerald (approved):
LGTM

nagisa (approved):
Thanks! This is wonderful.

PR #132, opened 2024-12-28
Expose `TRACY_NO_CRASH_HANDLER` define as default `crash-handler` feature flag

Mrmaxmeier:
While this is surely a _very_ niche feature, turning off the default crash handler can be useful in situations where the tracee sets up and uses signal handlers in normal operation.


nagisa:
This makes sense to me, thank you! Do you need a release with this right now or is it fine for you to depend on a git version for the time being?

Mrmaxmeier:
I'm fine with waiting for the next release. Thanks!

PR #144, opened 2025-08-29
Fibers PoC

torsteingrindvik:
It's not pretty (the Rust code I mean)- I just wanted to see what it looked like.

image

_EDIT: I whoopsed the location of the join prints, nvm them_

image

One strange thing is that it reports finishing after negative time:

image

The example follows section "3.10 Fiber" of the Tracy manual. Kinda. Added some subspans just to see things nest well.

All in all I think this is interesting- doing this properly could allow having Rust work sent across threads properly tracked. Not sure how it looks in a real life complex app with lots and lots of spans though.

Opening up in case someone wants to try/discuss. I don't intend to do anything more at the moment.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -14,7 +14,8 @@ Examples for using `tracy-client` in a wgpu application
 """
 
 [dependencies]
-tracy-client = { path = "../tracy-client" }
+tracy-client = { path = "../tracy-client", features = ["fibers"] }
+tracy-client-sys = { path = "../tracy-client-sys" }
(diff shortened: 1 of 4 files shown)


nagisa:
Thank you for experimenting with this. Fiber-based visualization is definitely desirable for Rust, especially async and tracing integration. Unfortunately the profiling perf impact simply by enabling fiber feature is so severe that I never really pursued it further. It definitely is not something that we could enable by default, and that alone requires a lot of design work, unfortunately.

PR #151, opened 2025-11-21
Allow (static) custom source locations for spans

### LINKED ISSUE #106: Support span colours
The new release of tracy now supports setting colours when allocating srcloc for span. We should too at least in `tracy-client`.

### LINKED ISSUE #108: Define custom source location for spans
We have a bunch of code that's generated by another language, and a build script creates the rust bindings for every method defined in the other language. It'd be really nice to be able to generate a `span_location!()` for each method, pointing to the true original source and line number, rather than it just assuming that `file!()`/`line!()` works.

The easiest would probably just be adding those as extra values to the macro, e.g. `span_location!(name, function, file, loc)`, but anything that works at compile time works for us. It'd get generated and stored in a giant `pub const FUNCTIONS = ....`.

moulins:
Closes #108 by expanding `span_location!` to allow overriding source locations:
```rs
let location: &'static SpanLocation = span_location! {
    name: "another name",
    function: "my::fake::function",
    file: "path/to/fake/file.rs",
    line: 42,
    color: 0x99CC66, // in RGB format
};
```

As an optimization, when the `function` name is provided statically, the `Lazy` wrapper is skipped.

EDIT: Also closes #106 by adding color support to static spans

-----

I considered adding a `const SpanLocation::new` constructor, but this would mean adding a way of making static nul-terminated strings outside of `FrameName/PlotName::new_leak`, so I skipped this for now. What do you think about adding an unified `StaticName/static_name!` type/macro that could be used everywhere Tracy expects such strings?

### DIFF
--- a/tracy-client/src/lib.rs
+++ b/tracy-client/src/lib.rs
@@ -74,12 +74,12 @@ pub mod internal {
             SpanLocation {
                 data: sys::___tracy_source_location_data {
                     name: span_name.cast(),
-                    function: function_name.as_ptr(),
+                    // Leak the string; `make_span_location` is only used inside `static`s.
+                    function: function_name.into_raw().cast(),
                     file: file.cast(),
                     line,
                     color: 0,
                 },
-                _function_name: function_name,
             }
         }
         #[cfg(not(feature = "enable"))]
--- a/tracy-client/src/lib.rs
+++ b/tracy-client/src/lib.rs
@@ -67,6 +67,7 @@ pub mod internal {
         span_name: *const u8,
         file: *const u8,
         line: u32,
+        color: u32,
     ) -> SpanLocation {
         #[cfg(feature = "enable")]
         {
@@ -78,7 +79,32 @@ pub mod internal {
                     function: function_name.into_raw().cast(),
                     file: file.cast(),
                     line,
-                    color: 0,
+                    color,
+                },
+            }
+        }
+        #[cfg(not(feature = "enable"))]
+        crate::SpanLocation { _internal: () }
+    }
+
+    #[inline(always)]
+    #[must_use]
+    pub const fn static_span_location(
+        function: *const u8,
+        name: *const u8,
+        file: *const u8,
+        line: u32,
+        color: u32,
+    ) -> SpanLocation {
+        #[cfg(feature = "enable")]
+        {
+            SpanLocation {
+                data: sys::___tracy_source_location_data {
+                    name: name.cast(),
+                    function: function.cast(),
+                    file: file.cast(),
+                    line,
+                    color,
                 },
             }
         }
--- a/tracy-client/src/span.rs
+++ b/tracy-client/src/span.rs
@@ -1,5 +1,4 @@
 use crate::{adjust_stack_depth, Client};
-use std::ffi::CString;
 
 /// A handle representing a span of execution.
 ///
@@ -19,8 +18,6 @@ pub struct Span {
 ///
 /// Construct with the [`span_location!`](crate::span_location) macro.
 pub struct SpanLocation {
-    #[cfg(feature = "enable")]
-    pub(crate) _function_name: CString,
     #[cfg(feature = "enable")]
     pub(crate) data: sys::___tracy_source_location_data,
     #[cfg(not(feature = "enable"))]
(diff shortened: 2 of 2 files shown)


Ralith:
> this would mean adding a way of making static nul-terminated strings

This crate predates it, but rust now has C-string literals which can be used for this.

moulins:
> This crate predates it, but rust now has C-string literals which can be used for this.

Oh, duh! Still, unlike a bare `&'static CStr`, a `StaticName` type has the advantage of being completely cfg'd out when the `"enable"` feature isn't set.

nagisa (commented):
Hm, I believe it should now be possible to construct a `\0` terminated byte arrays from strings without going through Lazy, even without forcing use of `c""` literals on them. Let me think about that a 'lil more.

tracy-client/src/span.rs
```diff
@@ -194,45 +191,86 @@ impl Drop for Span {
...
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __span_location_helper {
+    (
+        $([name $name: expr])?
+        [function $function: expr]
+        $([file $file: expr])?
+        $([line $line: expr])?
+        $([color $color: expr])?
```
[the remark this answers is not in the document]
moulins: Interesting, I'll have to think on how to do it while preserving the special-casing when defaulting the `function` field.

Ralith:
> a StaticName type has the advantage of being completely cfg'd out when the "enable" feature isn't set.

I don't think this is a meaningful advantage. This isn't a significant amount of space use, and the linker should do just fine at dropping unreferenced symbols anyway.

tracy-client/src/lib.rs
```diff
@@ -67,19 +67,45 @@ pub mod internal {
...
+        {
+            SpanLocation {
+                data: sys::___tracy_source_location_data {
+                    name: name.cast(),
+                    function: function.cast(),
+                    file: file.cast(),
+                    line,
+                    color,
                 },
-                _function_name: function_name,
```
nagisa, line 82: Although you're right that tracy requires function names to be effectively 'static, I think this change makes things strictly worse from the resilience and maintainability perspective.

While there is no practical use for having `SpanLocation` be dropped at all, I can imagine that in the future there may be use-cases that would drop `SpanLocation` (e.g. because it never got used actually, or maybe in some on-demand tracing scenarios...)

I think the first commit should be removed or adjusted to make the `CString` field an `Option`.

tracy-client/src/span.rs
```diff
@@ -194,45 +191,86 @@ impl Drop for Span {
...
+    ) => {{
+        static LOC: $crate::internal::SpanLocation =
+            $crate::internal::static_span_location(
+                concat!($function, "\0").as_ptr(),
```
nagisa, line 236: Lets have users supply `&'static CStr` (via `c"literals"`.) You can force this by doing a

```rust
CStr::as_ptr($argument)
```

SAme for when user provides a custom name, filename.

tracy-client/src/span.rs
```diff
@@ -194,45 +191,86 @@ impl Drop for Span {
...
+}
+
+#[doc(hidden)]
+#[macro_export]
+macro_rules! __span_location_helper {
+    (
+        $([name $name: expr])?
+        [function $function: expr]
+        $([file $file: expr])?
+        $([line $line: expr])?
```
nagisa, line 231: I feel like it makes little sense to allow overriding just the file, or just the line individually, but at the same time I have a hard time justifying why the library should prevent users from doing so :shrug:

nagisa (commented):
Having thought about it, I feel like trying to avoid the `Lazy` right now is making things more difficult than strictly necessary. If there is proof that it is causing a significant overhead, we can think about alternative approaches other than `span_location` potentially, but in absence of such proof (or in absence of *significant* impact) I'd rather wait until we get const-stable ways to obtain the function name instead.

Until then I think keeping `Lazy` will make it much more straightforward to provide a neat API. I reworked some code in https://github.com/nagisa/rust_tracy_client/pull/155 to that effect. Please take a look...

tracy-client/src/span.rs
moulins: Requiring C-string literals would make the macro different from all other literal-taking macros in the crate (`span!`, `plot_name!`, etc) so I'm not sure this is a good idea?

Ralith: +1, this seems like an unnecessary step backwards in ergonomics. The only upside I can see is if someone might need to pass a non-literal `&'static CStr`, but it still has to be a constant, so I'm not sure how that could arise.

PR #155, opened 2026-01-05
Custom locations prototype 2

nagisa:


tracy-client/src/span.rs
```diff
@@ -194,45 +191,73 @@ impl Drop for Span {
...
+                        line: line!(),
+                        color: 0,
+                    }
+                }.make()
             });
-        &*LOC
+        &LOC
     }};
+    (@internal_field_value function $value:expr) => {
+        $crate::internal::Cow::<'static, $crate::internal::CStr>::from($value)
```
nagisa, line 251: Salient question is whether we should allow `Cow` for other two stringy fields too.

